PR #23, opened 2018-05-24
fix rfc5766 typo

misi:


misi:
My time is limited but I will try to do my best.. 
Many Thanks for the kind words and for the invite. :smiley:

Issue #27, opened 2018-12-07
Too many open files bug


superp00t:
After running the server for a day and getting a regular influx of users, I was unable to use the TURN server and got many of these errors in the console: 

```
2018/12/07 17:32:57 Failed to handle ALLOCATE-REQUEST from x.x.x.x:37291: listen udp4 0.0.0.0:0: socket: too many open files
```

backkem:
The most likely problem seems to be that the `RelaySockets` are leaking. This won't be easy to debug without the ability to reproduce. Maybe the logs could contain a clue. If not, are you able to dig into the code to assist in debugging?

One thing I did notice is that a `Refresh` requests with a 0 duration seem to cause the `lifetimeTimer` to be stopped. That could cause the associated `RelaySocket` to never be closed. This is just a wild guess at this point thought.

Sean-Der:
Sorry this took so long @superp00t I am taking a break from working on `pion/webrtc` and want to spend some time on TURN.

PR #51, opened 2019-06-16
Improve channel_bind code coverage

songjiayang:
1. change ChannelBind's ID to turn.ChannelNumber.
2. move channelBindTimeout to StartArguments.

### DIFF
--- a/internal/allocation/channel_bind.go
+++ b/internal/allocation/channel_bind.go
@@ -4,29 +4,29 @@ import (
 	"fmt"
 	"net"
 	"time"
-)
 
-const channelBindTimeout = time.Duration(10) * time.Minute
+	"github.com/gortc/turn"
+)
 
 // ChannelBind represents a TURN Channel
 // https://tools.ietf.org/html/rfc5766#section-2.5
 type ChannelBind struct {
 	Peer          net.Addr
-	ID            uint16
+	ID            turn.ChannelNumber
 	allocation    *Allocation
 	lifetimeTimer *time.Timer
 }
 
-func (c *ChannelBind) start() {
-	c.lifetimeTimer = time.AfterFunc(channelBindTimeout, func() {
+func (c *ChannelBind) start(lifetime time.Duration) {
+	c.lifetimeTimer = time.AfterFunc(lifetime, func() {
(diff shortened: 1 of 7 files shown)


internal/allocation/channel_bind.go
```diff
@@ -4,15 +4,17 @@ import (
 	"fmt"
 	"net"
 	"time"
+
+	"github.com/gortc/turn"
 )
 
-const channelBindTimeout = time.Duration(10) * time.Minute
+var channelBindTimeout = time.Duration(10) * time.Minute
```
Sean-Der: Can we make this an argument to the server instead?

That way users can configure via https://github.com/pion/turn/blob/master/turn.go#L23 and then we can keep the default `10` minutes still!

Also would be great to have tests for this. I think we are/were leaking at one point, and we can close the ticket if you can assert that timeouts actually work/tear down everything :)

Sean-Der:
Fantastic improvement! Nice work :) 

Also if we make sure we are properly tearing down channels I think we can close out https://github.com/pion/turn/issues/27

internal/allocation/channel_bind.go
songjiayang: Put `channelBindTimeout ` into start configure params is a good idea, I will have a try.

songjiayang:
Hi @Sean-Der , review again please.

 I just add a attribute named `permissionLifetime `  to server. I think `ChannelBind` should have the same lifetime as `CreatePermission` lifetime, so it's 5 minutes for default.

internal/server/server.go
```diff
@@ -42,6 +43,7 @@ func NewServer(realm string, a AuthHandler) *Server {
 		authHandler:        a,
 		manager:            &allocation.Manager{},
 		reservationManager: &allocation.ReservationManager{},
+		permissionLifetime: 5 * time.Minute,
```
Sean-Der: Can we keep the value before `10 minutes`

beyond that LGTM! If you can change that will approve, thanks @songjiayang :)

songjiayang: @Sean-Der now move channelBindTimeout to StartArguments, and change it to 10 minutes with default.

songjiayang:
LGTM

PR #52, opened 2019-06-18
Support vnet

enobufs:
Relates to pion/webrtc#712

@Sean-Der Beside support of vnet and using pion/logger (instead of fmt.Print*), I made the following mildly major changes and wanted to draw early attention from you.

### Dropping IP level option as it is not portable (ReadFromCM => ReadFrom)
I guess the IP-level option (the "control message" - ipv4.DstFlag) was used because you thought real IP address other than 0.0.0.0 was necessary for five-tuple comparison.
I believe, if a specific IP address matters (e.g. generating relay address, etc), then we should have a dedicated socket for each local IP addresses instead of using "any-addr" (0.0.0.0). In most cases, we would only need one IP address (on eth0). This way, we can get away from having to use the any-addr.  The any-addr certainly makes things easier (you don't need to find your IP address) but price for needing IP-level option seems to be more expensive IMHO.

Let me spend a bit more time to understand how this affect fileTuple comparison. Either way though I think it is better to have `AddListener` method to allow listening on a multiple sockets in case the turn server has more than one IP address...

### Removed `turn.go` (the wrapper for client.Client and server.Server)
To support vnet, I need to access APIs (internal/client and internal/server). I found the API in the internal is pretty simple. So, I removed the wrapper, and moved server.go and client.go to the top (under name of `turn` package).   I modified cmd/simple-turn and cmd/client - as you can see, these are still simple enough I believe.


enobufs:
@ernado How does gortcd deal with the listening IP gathering?

ernado:
@enobufs
Via iterating over every interface and creating a listener for each one.

There is a config that is similar to "listening up".

Currently the NAT problem is not solved (if the interface IP differs from public IP it can break), I think that this should be solved via configuration (some kind of mapping)

enobufs:
@ernado Thanks! That's the way to go!
Yeah, the option for external (public) IP mapping (-X option for coturn) would be desired (AWS EC2, etc). Pion/turn should have that one too...

songjiayang:
@enobufs Nice work.

I have merged a pr https://github.com/pion/turn/pull/51 before, so there are some conflicting files need you to resolve.

enobufs:
Ok resolved all merge conflict with @songjiayang 's awesome allocation time refactor.  All tests passed - whew!   Merging now!

PR #53, opened 2019-06-19
Make server right close

songjiayang:
1. server close with defer.
2. make all allocation closeable when server close.

Maybe it can fix the bug https://github.com/pion/turn/issues/27 .

### DIFF
--- a/internal/allocation/allocation_manager.go
+++ b/internal/allocation/allocation_manager.go
@@ -2,6 +2,7 @@ package allocation
 
 import (
 	"fmt"
+	"log"
 	"net"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


Sean-Der:
I am not sure this is actually an issue anymore!

I have made some changes since #27 was filed, we should probably write some tests before we make any changes.

I think it should be pretty simple!
* Open 10 allocations (with a timeout of a minute)
* Measure how many sockets the current process has open

songjiayang:
@Sean-Der I will write some unit tests for this case.

Sean-Der:
I don't think we should land this PR (but please push back if you disagree)

I like that we return errors to users, logging is nice only if the error is async.

songjiayang:
@Sean-Der you are right,  maybe we can close server in`simple-turn/main.go`

PR #58, opened 2019-06-25
Refactor allocation lefetimeTimer handle function

songjiayang:
Add Manager close and Allocation timeout unit test.

### DIFF
--- a/internal/allocation/allocation_manager.go
+++ b/internal/allocation/allocation_manager.go
@@ -96,9 +96,7 @@ func (m *Manager) CreateAllocation(fiveTuple *FiveTuple, turnSocket net.PacketCo
 	a.RelayAddr = conn.LocalAddr()
 
 	a.lifetimeTimer = time.AfterFunc(lifetime, func() {
-		if err := conn.Close(); err != nil {
-			a.log.Errorf("Failed to close listener for %v", a.fiveTuple)
-		}
+		m.DeleteAllocation(a.fiveTuple)
 	})
 
 	m.lock.Lock()
--- a/internal/allocation/allocation.go
+++ b/internal/allocation/allocation.go
@@ -166,6 +166,8 @@ func (a *Allocation) Close() error {
 	}
 	close(a.closed)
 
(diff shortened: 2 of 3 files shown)


internal/allocation/allocation_manager.go
```diff
@@ -83,7 +83,7 @@ func (m *Manager) CreateAllocation(fiveTuple *FiveTuple, turnSocket ipnet.Packet
 	a.RelayAddr = listener.LocalAddr()
 
 	a.lifetimeTimer = time.AfterFunc(lifetime, func() {
-		if err := listener.Close(); err != nil {
```
Sean-Der: Nice catch. Right now we don't ever clean up allocations (but we close the socket) Does this change make the tests now pass?

I think we leak the `listener` now though, we should probably put the cleanup in `(a *Allocation) Close()`

songjiayang: I think `(a *Allocation) Close()` has include `listener` close logic, see the code [allocation.go#L186](https://github.com/pion/turn/blob/master/internal/allocation/allocation.go#L186). 

Change to `listener.Close()` , the tests pass yet because  the Allocation will be deleted in `(a *Allocation) packetHandler`, details to see [allocation.go#L217](https://github.com/pion/turn/blob/master/internal/allocation/allocation.go#L217)

Sean-Der: Should we leave this line unchanged then? Approving the PR but w/e you think is best go for it :) 

Maybe we can refactor the entire shut down/destroy logic!

enobufs:
@songjiayang #52 has been merged! Hope the conflict wouldn't be too bad for you. We can not live without your tests and love for turn! :) - if you need help, let me know!

internal/allocation/allocation_manager.go
songjiayang: @Sean-Der  Just change `Allocation.lifetimeTimer` handle function to `m.DeleteAllocation()`, I think it's more clear, code review again please.

PR #59, opened 2019-06-27
Change manager allocations type

songjiayang:
From slice to map.

### DIFF
--- a/internal/allocation/allocation.go
+++ b/internal/allocation/allocation.go
@@ -209,9 +209,7 @@ func (a *Allocation) packetHandler(m *Manager) {
 	for {
 		n, srcAddr, err := a.RelaySocket.ReadFrom(buffer)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


songjiayang:
@enobufs yes, I can create branch on pion/turn, do it at next pr.

Sean-Der:
Oh yea fantastic change @songjiayang you are doing amazing work with `pion/turn` really excited to see this.

I was talking to @cameron-elliott today in person and he threw out the idea of really ripping out the guts of pion/turn of removing the needs for all the locking and shooting for high performance.

It would be really interesting to see if we could beat the performance of coturn if we really pushed it with optimizations. Also explore interesting things like DPDK.

I don't really have the bandwidth to push this project, but I would love to see something like that happen :) as long as all the tests pass/we are tracking performance I think Go gives us the chance to make something really amazing!

PR #63, opened 2019-06-30
[WIP] Change allocation binds data type

songjiayang:
1. Add channelNumberBindings to store the relationship with binding numbers and ChannelBind.
2. Add channelPeerBindings to store the relationship with binding peer's addr and ChannelBind.
3. Fix the bug about invalid ChannelBind addr in allocation's packageHandler.

### DIFF
--- a/internal/allocation/allocation.go
+++ b/internal/allocation/allocation.go
@@ -9,35 +9,40 @@ import (
 	"github.com/gortc/turn"
 	"github.com/pion/logging"
 	"github.com/pion/stun"
-	"github.com/pion/turn/internal/ipnet"
 	"github.com/pkg/errors"
 )
 
 // Allocation is tied to a FiveTuple and relays traffic
 // use CreateAllocation and GetAllocation to operate
 type Allocation struct {
-	RelayAddr           net.Addr
-	Protocol            Protocol
-	TurnSocket          net.PacketConn
-	RelaySocket         net.PacketConn
-	fiveTuple           *FiveTuple
-	permissionsLock     sync.RWMutex
-	permissions         map[string]*Permission
-	channelBindingsLock sync.RWMutex
-	channelBindings     []*ChannelBind
-	lifetimeTimer       *time.Timer
-	closed              chan interface{}
-	log                 logging.LeveledLogger
+	RelayAddr   net.Addr
+	Protocol    Protocol
+	TurnSocket  net.PacketConn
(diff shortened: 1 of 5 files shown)


internal/allocation/allocation.go
```diff
@@ -9,35 +9,40 @@ import (
...
+	Protocol    Protocol
+	TurnSocket  net.PacketConn
+	RelaySocket net.PacketConn
+	fiveTuple   *FiveTuple
+
+	permissionsLock sync.RWMutex
+	permissions     map[string]*Permission
+
+	channelBindingsLock   sync.RWMutex
+	channelNumberBindings map[turn.ChannelNumber]*ChannelBind
```
Sean-Der: Are we storing information twice for performance reasons?

I would like to avoid if possible, just to avoid possible bugs in the future where people modify things incorrectly.

Sean-Der: I would be curious how often `GetChannelByID` and `GetChannelByAddr` is called. We can pick the best storage off of that.

If they are equal maybe we can store twice, but does feel wrong. But not a huge deal!

internal/allocation/allocation.go
```diff
@@ -80,70 +85,50 @@ func (a *Allocation) RemovePermission(fingerprint string) {
...
+		a.channelNumberBindings[c.Number] = c
+		a.channelPeerBindings[c.Peer.String()] = c
+		a.channelBindingsLock.Unlock()
 
-		c.allocation = a
-		a.channelBindings = append(a.channelBindings, c)
 		c.start(lifetime)
-
-		// Channel binds also refresh permissions.
-		a.AddPermission(NewPermission(c.Peer, a.log))
```
Sean-Der: Why did we remove `a.AddPermission(NewPermission(c.Peer, a.log))`, I don't remember this code at all anymore :/

songjiayang: I think  `Permission` and `ChannelBind` should create with different API rather than auto create Permission by ChannelBind. I can't find this logic in RFC 5766, Maybe there is something I missed.

internal/allocation/allocation.go
songjiayang: Sending message with channel once,  caller will call `GetChannelByID ` method  ( or callee (peer) call `GetChannelByAddr ` ) once.  I think the most important thing is that how many the binding channels for an `Allocation` will be,  100 or 1000?

PR #64, opened 2019-07-02
[WIP] Refactor code with context

songjiayang:
To make handler's code more simple and clear.

Hi @Sean-Der  @enobufs , I  do some changes about the server handler,  please give me some suggestion about this refactor, if it's ok, I will complete the unit test.

ps: I will add unit test to coverage all the changes before merged.


context.go
```diff
@@ -0,0 +1,111 @@
...
+	if err := assertMessageIntegrity(ctx.msg, ourKey[:]); err != nil {
+		return ctx.ResponseWithError(err, stun.CodeBadRequest)
+	}
+
+	ctx.messageIntegrity = stun.NewLongTermIntegrity(usernameAttr.String(), realmAttr.String(), password)
+
+	return nil
+}
+
+// Response is a help method to send success response.
```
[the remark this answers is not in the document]
enobufs: Oops, it actually sends the response.. then it should be a verb. `Respond` and `RespondWIthError` like the original code was.

context.go
```diff
@@ -0,0 +1,111 @@
...
+		attrs...,
+	); sendErr != nil {
+		err = errors.Errorf(strings.Join([]string{sendErr.Error(), err.Error()}, "\n"))
+	}
+
+	return err
+}
+
+// Return is a help method to send response message.
+func (ctx *Context) Return(class stun.MessageClass, attrs ...stun.Setter) error {
```
enobufs: Would this method need to be public?
The name 'return' does not clearly represent what it does...  can we rename it as 'buildAndSend'?

enobufs:
I generally like the idea of introducing a context (of each turn protocol transactions). This kind of refactor makes me a bit nervous particularly when I am not very familiar with TURN protocol. @Sean-Der @ernado any thoughts?

context.go
```diff
@@ -0,0 +1,111 @@
...
+
+	messageIntegrity stun.MessageIntegrity
+}
+
+// NewContext creates a new instance of Context.
+func NewContext(conn net.PacketConn, srcAddr net.Addr, m *stun.Message, realm string, authHandler AuthHandler) *Context {
+	return &Context{
+		conn:        conn,
+		srcAddr:     srcAddr,
+		dstAddr:     conn.LocalAddr(),
```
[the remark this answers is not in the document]
songjiayang: removed

context.go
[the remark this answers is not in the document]
songjiayang: changed, thanks

context.go
songjiayang: rename to buildAndSend

songjiayang:
> I have left some feedback. When you have time, please review them.

Hi @enobufs  thanks you so much, I have changed the code with your review, check again pls.

Sean-Der:
@songjiayang @enobufs 

I agree with this refactor! Yea, it is a little scary with how little code coverage I provided :( but as long as we cover everything/test we can feel good about it in the future :)

context.go
```diff
@@ -0,0 +1,109 @@
...
+	"crypto/md5" // #nosec
+	"net"
+	"strings"
+
+	"github.com/pion/stun"
+	"github.com/pkg/errors"
+)
+
+// Context is used to hold the context of a request
+type Context struct {
```
Sean-Der: Does `Context` need to be public?

context.go
```diff
@@ -0,0 +1,109 @@
...
+	msg     *stun.Message
+
+	realm       string
+	authHandler AuthHandler
+
+	messageIntegrity stun.MessageIntegrity
+}
+
+// NewContext creates a new instance of Context.
+func NewContext(conn net.PacketConn, srcAddr net.Addr, m *stun.Message, realm string, authHandler AuthHandler) *Context {
```
Sean-Der: If possible I would try to push everything into `internal` possible. Just so if people jump into `pion/turn` and want to read it can make it as easy as possible.

Sean-Der:
Just did a quick review @songjiayang, this looks great to me! My only asks.

* Can we move `context` and probably `turn.go` into internal somehow? I just really want `/` to be as minimal as possible.


As a nice follow up the tests with `pion/turnc` would be amazing :) but don't worry about it yet! I don't want to burn you out w/e you think the best path forward for `pion/turn` is I think we should follow. @enobufs and I are busy in other places, so the most important thing is that you can drive the project/have ownership how you think is best :)

turn.go
```diff
@@ -262,63 +172,56 @@ func (s *Server) handleAllocateRequest(conn net.PacketConn, srcAddr net.Addr, m
...
+		return errors.Errorf("No allocation found for %v:%v", ctx.srcAddr, ctx.conn.LocalAddr())
 	}
 
-	messageIntegrity, _, err := authenticateRequest(curriedSend, m, stun.MethodCreatePermission, s.realm, s.authHandler, srcAddr)
-	if err != nil {
+	if err := ctx.Authenticate(); err != nil {
 		return err
 	}
+	if ctx.messageIntegrity != nil {
+		return nil
```
Sean-Der: Should we return a `non-nil` error here? Just so when debugging we know `Authenticate` didn't give us a value.

songjiayang: Cool, I think `non-nil` error here is better.

context.go
songjiayang: Can't agree and more, thanks.

songjiayang: I add `server` and `client` sub package to `internal ` and move `context.go`  to `server`  package.

context.go
songjiayang: fixed.

enobufs:
@songjiayang I saw your commit at 83702c9, that you are moving server and client entirely back into internal folder.  You cannot do it because APIs under `internal` is not accessible from outside the package. (go will complain if you try to access it from outside pion/turn.

What @Sean-Der meant was to move what were in turn.go because stuff in turn.go were not public APIs. I actually attempted to do it but that also had a problem because the private methods references `Server` (under turn package),  which would create circular dependencies, and go will complain it as well...

Only thing we could do is, modify methods in turn.go file not to depend on Server class (a collection of simple methods (not property methods), then move them under internal (just like allocation stuff are). If that is difficult, then we should leave them at the top level.

@Sean-Der what do you think? I personally feel leaving turn.go as it is now is fine, mostly because those are still private methods. (users can't use them). You can still see files even that are inside internal as well...

songjiayang:
@enobufs Thanks for your suggestion and moved `client` and `server` to project's root folder. I support the idea of @Sean-Der , we need to make the root folder simple.

Issue #65, opened 2019-07-03
TURN server should behave 'address-restricted-filtering'


enobufs:
### Your environment.
  - Version: v1.2.0
  - Browser: n/a

### What did you do?
I wrote a test: https://github.com/pion/ice/blob/topic-vnet/connectivity_vnet_test.go
in which there are symmetric NATs on both ends, if one end did not have TURN server URL (but has STUN server URL), the connection fails! 

### What did you expect?
srflx-relay pair should succeed, when both ends have a symmetric NAT, and only one end have TURN server URL.

### What happened?
Connection fails.
Looking at [this spec](https://tools.ietf.org/html/rfc5766#section-2.3), TURN server (Permission) should do `address-restricted filtering`
```
To ease concerns amongst enterprise IT administrators that TURN could be used to bypass corporate firewall security, TURN includes the notion of permissions. TURN permissions mimic the address-restricted filtering mechanism of NATs that comply with [RFC4787].
```

But looking at the current implementation of TURN permission, the fingerprint includes port number...  We should exclude the port number from the fingerprint and its comparison with the srcAddr.

@songjiayang Can you confirm my above observation?

enobufs:
Looking at coturn, its permission hashtable only evaluate IP address/family, not port: https://github.com/coturn/coturn/blob/master/src/server/ns_turn_allocation.c#L248

We should do the same!

songjiayang:
@enobufs You are right. With RFC 5766, `Permission`  only include IP and `ChannelBind` with IP and port.  I have created  the same issue https://github.com/pion/turn/issues/41 for this problem.

PR #68, opened 2019-07-03
Use IP address only for permission fingerprint

### LINKED ISSUE #41: Permission should not include port
From the rfc https://tools.ietf.org/html/rfc5766#section-2.3, it should only include the IP address and lifetime.

enobufs:
Please see #65 for details

NOTE:
* #41 is a duplicate

Fixes #41 and #65

### DIFF
--- a/internal/allocation/allocation.go
+++ b/internal/allocation/allocation.go
@@ -30,6 +30,16 @@ type Allocation struct {
 	log                 logging.LeveledLogger
 }
 
+func addr2Fingerprint(addr net.Addr) string {
+	switch a := addr.(type) {
+	case *net.UDPAddr:
+		return a.IP.String()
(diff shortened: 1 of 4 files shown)


enobufs:
@songjiayang I am just wondering if you subscribe pion's slack channel in Gopher workspace.  I noticed today that you were not in the slack channel... Lot's discussion was made today around TURN actually.

internal/allocation/allocation.go
```diff
@@ -30,6 +30,16 @@ type Allocation struct {
 	log                 logging.LeveledLogger
 }
 
+func addr2Fingerprint(addr net.Addr) string {
```
songjiayang: How about rename to `addr2IPFingerprint `

Issue #56, opened 2019-06-23
Support of external IP address mapping


enobufs:
## Summary
When you want to host TURN server on AWS EC2 for instance, the instance can have a private IP address, but not public address directly. It allows us to have an associated public IP address (or Elastic IP, etc, there will be a NAT in betwee), but TURN server does not know about the public IP address at all. Since it is TURN server's responsibility to report the public IP address to the TURN client, there need to be a way to tell the associated public IP address to the TURN server.

## Motivation
To make pion/turn available for real use cases.

## Describe alternatives you've considered
Tell the external IP address via:
1. command line option
1. environment variable

## Additional context
coturn, for instance, have -X option for this purpose.

enobufs:
@Sean-Der Can we add this to the 2.2.0 project? If this, plus #82 is supported, I'd switch my TURN server to pion/turn. (how do I add this ticket to the project?)

PR #70, opened 2019-07-03
Use specified IP for relay socket

### LINKED ISSUE #69: Relay socket should not listen on 0.0.0.0
### Your environment.
  - Version: v1.2.0
  - Browser: n/a

### What did you do?
This bug has been introduced since we removed ipv4.FlagDst.
Turn protocol listen ports are now using "specified" IP addresses, but relay ports have not been taken care of.

### What did you expect?
Obviously, if listen addr is 0.0.0.0, permission search would fail because the permission hash is based on "specified" (non-any-addr) address.

### What happened?
Permission check was always failing because the IP address did not match the incoming packets source IP.

Fixes are incoming!

enobufs:
Fixes #69

See #69 for details.

I have added AddRelayIPAddr() method also. If this method is not used, then it will default to listening IPs.

#### Limitation/TODO:
You may add multiple relay IPs, but this fix only uses the first relay IP address. This limitation will need to be fixed by the time when we allow IPv6 or other transport protocols.

### DIFF
--- a/server.go
+++ b/server.go
@@ -52,6 +52,7 @@ type Server struct {
 	lock               sync.RWMutex
 	listeners          []*listener
 	listenIPs          []net.IP
+	relayIPs           []net.IP
 	listenPort         int
 	realm              string
 	authHandler        AuthHandler
@@ -100,6 +101,8 @@ func NewServer(config *ServerConfig) *Server {
 }
 
 // AddListeningIPAddr adds a listening IP address.
+// If not specified, it will automatically assigns the listening IP addresses
+// from the system.
 // This method must be called before calling Start().
 func (s *Server) AddListeningIPAddr(addrStr string) error {
 	ip := net.ParseIP(addrStr)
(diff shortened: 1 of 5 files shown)


turn.go
```diff
@@ -212,7 +212,11 @@ func (s *Server) handleAllocateRequest(conn net.PacketConn, srcAddr net.Addr, m
 	// Check current usage vs redis usage of other servers
 	// if bad, redirect { stun.AttrErrorCode, 300 }
 	lifetimeDuration := allocationLifeTime(m)
-	a, err := s.manager.CreateAllocation(fiveTuple, conn, requestedPort, lifetimeDuration)
+	a, err := s.manager.CreateAllocation(
+		fiveTuple, conn,
+		s.relayIPs[0], // TODO: allow more than one relay IP
```
songjiayang: I think `conn.LocalAddr`'s IP is ok here. But I have another question, if the turn server have multiple public `IP`， how to get the correspondence between public IP and intranet IP.

server.go
```diff
@@ -114,13 +117,37 @@ func (s *Server) AddListeningIPAddr(addrStr string) error {
...
 
+// AddRelayIPAddr adds a listening IP address.
+// Note: current implementation can have only one relay IP address.
+// If not specified, it will automatically assigns the relay IP addresses
+// from the system.
+// This method must be called before calling Start().
+func (s *Server) AddRelayIPAddr(addrStr string) error {
+	ip := net.ParseIP(addrStr)
+	if ip.To4() == nil {
+		return fmt.Errorf("Non-IPv4 address is not supported")
```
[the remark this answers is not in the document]
enobufs: Hmm... :) There are differing opinions there... Some told me to use either fmt.Errorf or errors package. errors is another package... :p To me, whichever.

turn.go
enobufs: Applications like TURN server or "routers" may have the bandwidth bottleneck at network interfaces (NIC, Ethernet, etc), and it is pretty common to have separate interfaces (just like routers, switches have multiple ports). So, I think it is a good idea to allow having TURN (listening) ports and relay ports.

CoTurn for instance provides that option also. See [turnserver.conf](https://github.com/coturn/coturn/blob/master/examples/etc/turnserver.conf)

server.go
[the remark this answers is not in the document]
enobufs: Having said that, you are the major contributor of pion/turn now.  ;) I can switch to errors.  Thanks for all your help.  (I am trying to fix connectivity issue for my project right now and that's why I am working on TURN. Hope it gets resolved soon!)

turn.go
songjiayang: Cool, I see the configuration `external-ip=60.70.80.92/172.17.19.102` can solve this problem.

server.go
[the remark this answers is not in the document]
songjiayang: In fact `fmt. Errorf`  and `errors.New` both are ok , it's a personal habit 😄.  PR approve now.

turn.go
enobufs: For that, I filed an issue as #56.
These are the ones I was talking about for separate relay IP options:
```
#relay-ip=172.17.19.105
#relay-ip=2607:f0d0:1002:51::5
```
FYI

PR #73, opened 2019-07-10
Make SOFTWARE attribute configurable

### LINKED ISSUE #18: Add support for customizing SOFTWARE attribute value
Currently this is hard-coded and should be specifiable either through configuration or environment variables.

namreg:
#### Description
Added ability to configure `SOFTWARE` attribute. For a client is configurable through command argument `software`, for a server - through environment variable`SOFTWARE`.
#### Reference issue
Fixes #18

### DIFF
--- a/sender.go
+++ b/sender.go
@@ -9,10 +9,8 @@ import (
 // Sender is responsible for building a message and sending it to the given addr.
 type Sender func(conn net.PacketConn, addr net.Addr, attrs ...stun.Setter) error
 
-// DefaultSender is a sender default implementation.
-var DefaultSender Sender = buildAndSend
-
-func buildAndSend(conn net.PacketConn, dst net.Addr, attrs ...stun.Setter) error {
+// DefaultBuildAndSend is a sender default implementation.
+func DefaultBuildAndSend(conn net.PacketConn, dst net.Addr, attrs ...stun.Setter) error {
 	msg, err := stun.Build(attrs...)
 	if err != nil {
 		return err
--- a/sender.go
+++ b/sender.go
@@ -9,8 +9,8 @@ import (
 // Sender is responsible for building a message and sending it to the given addr.
 type Sender func(conn net.PacketConn, addr net.Addr, attrs ...stun.Setter) error
 
-// DefaultBuildAndSend is a sender default implementation.
-func DefaultBuildAndSend(conn net.PacketConn, dst net.Addr, attrs ...stun.Setter) error {
+// defaultBuildAndSend is a sender default implementation.
+func defaultBuildAndSend(conn net.PacketConn, dst net.Addr, attrs ...stun.Setter) error {
 	msg, err := stun.Build(attrs...)
 	if err != nil {
 		return err
--- a/sender.go
+++ b/sender.go
@@ -1,0 +1,22 @@
+package turn
+
+import (
+	"net"
+
+	"github.com/pion/stun"
+)
+
+// Sender is responsible for building a message and sending it to the given addr.
+type Sender func(conn net.PacketConn, addr net.Addr, attrs ...stun.Setter) error
+
+// DefaultSender is a sender default implementation.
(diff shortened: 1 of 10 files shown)


Sean-Der:
Really fantastic work @namreg, this is great!

Would you mind adding a simple test just to confirm that `software` is added to requests/responses. You don't have to check everything, but just to make sure someone doesn't break your code in the future :)

thank you!

namreg:
@Sean-Der I've added some tests :)

Sean-Der:
Fantastic work, the tests look great!

sender.go
```diff
@@ -0,0 +1,20 @@
+package turn
+
+import (
+	"net"
+
+	"github.com/pion/stun"
+)
+
+// Sender is responsible for building a message and sending it to the given addr.
+type Sender func(conn net.PacketConn, addr net.Addr, attrs ...stun.Setter) error
```
Sean-Der: Oh that is really interesting! I bet that could be helpful if you want to insert a custom attr/inspect in every message. Nice idea.

sender.go
```diff
@@ -0,0 +1,20 @@
...
+
+// DefaultBuildAndSend is a sender default implementation.
+func DefaultBuildAndSend(conn net.PacketConn, dst net.Addr, attrs ...stun.Setter) error {
```
Sean-Der: Does this need to be public?

Sean-Der:
Fantastic work, the tests look great! Mind making that one visibility change and I will merge :)

If you are interested we would love to have you in [Slack](https://pion.ly/slack), we talk about a lot of fun P2P/WebRTC stuff there!

sender.go
namreg: Thank you! Yeah, it will allow us to analyze in-depth every message.

sender.go
namreg: No, no needs) fixed

namreg:
> If you are interested we would love to have you in Slack, we talk about a lot of fun P2P/WebRTC stuff there!

I am already there, thanks! :)

Sean-Der:
Fantastic work @namreg!

Thanks for the change, and I look forward to anything in the future :)

enobufs:
@namreg 
I am working on major changes to pion/turn Client implementation, and I noticed you have added `Sender` to its config parameter, and I am trying to understand why it was necessary... It appears to me that it has nothing to do with adding Software attribute.

See #75 

@Sean-Der

Sean-Der:
@enobufs 

Another thing that was useful was tracing all the outbound messages. I think @namreg is logging/analyzing everything that is being sent.

I think it would be ok to cut down the amount of arguments/make the API simpler as long as we can still do that!

enobufs:
@Sean-Der 

If that's the purpose, then I would approach in a different way...  The `Sender` is taking over the entire (original) "buildAndSend" including 'conn'!  ([the code](https://github.com/pion/turn/blob/f0df9a6e8c7e52afd24155f555f8384d6e9cb8f8/client.go#L76) I think it is too intrusive for our API.

To monitor what was sent, then I would provide a pipe of net.PacketConn to tap in what was sent.

If @namreg wanted to have complete flexibility in building STUN request (e.g. adding arbitrary attributes, etc), then with the new API, we have `client.PerformTransaction` method to use.

For now, I removed Sender option from the client config in my PR #75 during the rebase I have just did. (Server still has it but I think we should remove it later and provide another way.)

namreg:
@enobufs The main reason to add Sender was making code mode testable because I wanted to test code I have written.

> To monitor what was sent, then I would provide a pipe of net.PacketConn 

I haven't look at your PR yet, but I think `net.PacketConn` is not a very good abstraction - it is supposed using the only udp connection. But according to the rfc (also we have a corresponding issue #1 ) client and server can talk via TCP (+ TCP over TLS).

enobufs:
When TCP is introduced, we will likely add net.Conn for TCP connection.

PR #75, opened 2019-07-11
TURN client refactor

### LINKED ISSUE #74: TURN client refactor
### Issues:
* Current pion/turnc behaves as “port-restricted” NAT
* If the other end is a symmetric NAT, ICE cannot find a valid candidate pair
* Current pion/turnc is overly complex to fix this

### Related Issues on Github
* Switch to "address-restricted" permission (pion/turnc#11)
* Add vnet support (pion/ice#46)
* Share one UDP socket with all UDP candidates (pion/ice#76)

## Motivation
We will try to implement a new TURN client in a way that would solve all the above problems, without having to modify pion/turnc (would be hard).
The pion/turn has its own client but it does not support TURN protocol. It's a clean place to try out the new implementation of TURN client, while leaving the pion/turnc as is to leave us "two-way door".

Modify pion/turnc .... I have tried, but it is really complex. Abstraction of channel as a net.Conn has been found fundamentally intervening in allowing ICE to detect prflx candidate which is necessary to traverse remote symmetric NAT even local end uses TURN.

## Additional context
Details of issues I have found and my refactoring plan is described [here](https://docs.google.com/document/d/1kNHqKI3OL9CeoVJve7ehmWvOSj98G4lQLNubQvTYdpc)

enobufs:
Resolves #74

#### Description
Please read the details in #74


client.go
```diff
@@ -1,82 +1,235 @@
...
+}
+
+// Listen will have this client start listening on the conn provided via the config.
+// This is optional. If not used, you will need to call HandleInbound method
+// to supply incoming data, instead.
+func (c *Client) Listen() error {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	if c.conn == nil {
```
Sean-Der: Can we do 
```
if c.conn != nil {
    return fmt.Errorf("already listening")
}
```

Or is there a case `c.isListening == true && c.conn != nil` ? Don't know if helpful for testing!

Sean-Der: I would just not allow `c.conn == nil`, I would force the user to always pass the listener themselves (if possible)?

client.go
```diff
@@ -1,82 +1,235 @@
...
+		}
+		c.conn = conn
 	}
 
-	packet := make([]byte, 1500)
-	c.log.Debug("wait for STUN response...")
-	size, _, err := c.conn.ReadFrom(packet)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed to read packet from udp socket")
+	if c.isListening {
```
Sean-Der: I would put `isListening` inside a function (and personally would prefer an atomic) then we don't have to worry about taking/releasing a mutex!

client.go
```diff
@@ -1,82 +1,235 @@
...
+	return nil
+}
+
+// Close closes this client
+func (c *Client) Close() {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
+
+	if c.isListening {
+		c.conn.Close() // nolint:errorcheck
```
Sean-Der: Can we return this to the user, we can just put at the end of the function
```
if c.isListening {
  return c.conn.Close()
}

return nil
```

internal/client/channel.go
```diff
@@ -0,0 +1,15 @@
+package client
+
+import (
+	"sync"
+
+	"github.com/gortc/turn"
```
Sean-Der: Can we pull `turn.ChannelNumber`, just so we don't have any dependencies we don't control!

internal/client/conn.go
```diff
@@ -0,0 +1,305 @@
+package client
+
+import (
+	"fmt"
+	"io"
+	"math"
+	"net"
+	"time"
+
+	"github.com/gortc/turn"
```
Sean-Der: if possible remove `gortc` dependency. I am just worried about not being able to fix things outside our control.

client.go
```diff
@@ -87,11 +240,252 @@ func (c *Client) SendSTUNRequest(serverIP net.IP, serverPort int) (net.Addr, err
...
+
+	trRes, err = c.PerformTransaction(msg, c.turnServ, false)
+	if err != nil {
+		return nil, err
+	}
+	res = trRes.Msg
+
+	if res.Type.Class == stun.ClassErrorResponse {
+		var code stun.ErrorCodeAttribute
+		if err = code.GetFrom(res); err == nil {
```
Sean-Der: Possibly useful to include in `fmt.Errorf`

client.go
```diff
@@ -87,11 +240,252 @@ func (c *Client) SendSTUNRequest(serverIP net.IP, serverPort int) (net.Addr, err
...
+
+	var err error
+
+	switch {
+	case stun.IsMessage(data):
+		err = c.handleSTUNMessage(data, from)
+	case turn.IsChannelData(data):
+		err = c.handleChannelData(data)
+	default:
+		return false, nil // unhandled
```
Sean-Der: Might be worth a trace/debug message

client.go
```diff
@@ -87,11 +240,252 @@ func (c *Client) SendSTUNRequest(serverIP net.IP, serverPort int) (net.Addr, err
...
+func (c *Client) SendBindingRequest() (net.Addr, error) {
+	if c.stunServ == nil {
+		return nil, fmt.Errorf("STUN server address is not set for the client")
+	}
+	return c.SendBindingRequestTo(c.stunServ)
+}
+
+// Allocate sends a TURN allocation request to the given transport address
+func (c *Client) Allocate() (net.PacketConn, error) {
+	if c.relayedConn != nil {
```
Sean-Der: I would put a mutex on `Allocate`. A bad user could call Allocate twice, always assume the worst :)

internal/client/conn.go
enobufs: Yes, I agree and that's the goal as we discussed.
I am thinking about taking two steps.

1. Redesign TURN client in pion/turn (this issue)
1. Remove dependencies on gortc/*

Even in the first step, I am trying to **only** depend on structures and STUN message builders/parsers defined in gortc/turn. Once the first step is complete and made it functional, I will work on step two.  (I really need to fix this TURN issue asap for my project..)

internal/client/conn.go
```diff
@@ -0,0 +1,305 @@
...
+
+		case <-c.readTimer.C:
+			return 0, nil, &net.OpError{
+				Op:   "read",
+				Net:  c.LocalAddr().Network(),
+				Addr: c.LocalAddr(),
+				Err:  newTimeoutError("i/o timeout"),
+			}
+
+		case <-c.closeCh:
```
Sean-Der: Can we return an error here, are you worried there might be more data to read?  If so I would vote for just another read call (just to be explicit). I had to read this a few times to make sure I didn't miss anything!

internal/client/channel.go
enobufs: The type is actually associated with methods for message building/parsing. I'd like to the dependency removal after landing this TURN client's functional issue.

Sean-Der:
Amazing work @enobufs this code was a joy to read, really impressive how quickly you are getting this done.

This will close out a lot of TURN issues, I really appreciate you taking this on!

client.go
enobufs: Yes, I will add a debug message. (i will implement channel binding soon)

client.go
enobufs: Ah, that's true...  I'd need to review entire code for thread-safety.

internal/client/conn.go
enobufs: I can't remember why I added the closeCh... (I copied this code from vnet) Let me review.

internal/client/channel.go
Sean-Der: ACK, sounds good to me!

internal/client/conn.go
Sean-Der: Agree!

client.go
enobufs: Now, setting conn is required. NewClient() will fail otherwise.

client.go
enobufs: Now, conn is provided from outside. It is turn.Client's responsibility to close the conn. (isListening is no longer needed)

client.go
enobufs: That was my typo.  I have added the log correctly.

client.go
enobufs: Added!

internal/client/conn.go
enobufs: closeCh was necessary because it was not safe to close readCh, if I remember correctly. But, there's anoteher problem. User may call ReadFrom() after the conn is closed. At that point, it may happen to read data again (if it is the buffer), which is not the expected behavior.  For that, I have added a atomic flag `closed` (I introduced `AtomicBool`) in order to return error when that happens.

client.go
enobufs: The solution wouldn't be the mutex... After some thoughts, I introduced a simple try-lock, which would return an error immediately when Allocate() was called while another goroutine's Allocate() call is not complete yet. Can you review it again?

enobufs:
Refresh for allocation and permissions are done. What's left is ChannelBind/ChannelData!

client.go
enobufs: conn is no longer closed by the client.

client.go
```diff
@@ -87,11 +198,269 @@ func (c *Client) SendSTUNRequest(serverIP net.IP, serverPort int) (net.Addr, err
...
+		err = c.handleSTUNMessage(data, from)
+	case turn.IsChannelData(data):
+		err = c.handleChannelData(data)
+	default:
+		// assume, this is an application data
+		c.log.Tracef("non-STUN/TURN packect, unhandled")
+		return false, nil // unhandled
+	}
+
+	if err != nil {
```
Sean-Der: I would always `return true, err` (if err is nil we are returning the same thing)

internal/client/transaction.go
```diff
@@ -0,0 +1,156 @@
...
+
+	"github.com/pion/stun"
+)
+
+const (
+	maxRtxInterval time.Duration = 1600 * time.Millisecond
+)
+
+// TrResult is a bag of result values of a transaction
+type TrResult struct {
```
Sean-Der: Maybe `TransactionResult` is better, just to remove any ambiguity.

internal/client/conn.go
Sean-Der: Nice! Didn't think of that.

Do we need the `break loop`? The `for` loop should run again and bail I believe?

Sean-Der:
> Refresh for allocation and permissions are done. What's left is ChannelBind/ChannelData!

PR looks fantastic! Only minor nits.

After this release I am going to take a crack at performance stuff next. Especially with us putting debug/print statements everywhere that worries me a little bit. Wish it was compile time a bit :/

client.go
enobufs: ah ok, will do

internal/client/transaction.go
enobufs: True. I will use TransactionResult.

internal/client/conn.go
enobufs: Oh true. Will remove the break/loop.

enobufs:
Besides the details, I guess you are ok with the overall architecture..(?)  I have some ideas on how to integrate this new client to pion/ice, but I am not fully confident about my understanding in how pion/ice is implemented. (I should be able to start working on the integration into pion/ice.)

## Simplified CandiateRealy
With the net.PacketConn interface, we may be able to unify the candidate base. CandidateRelay will be a lot simpler 

## Demultiplexing
This relates to pion/ice#76. All candidate should share their the same "base". We will need to implement how to demultiplex the inbound packet from the same conn. I have a rough sketch in [this doc](https://docs.google.com/document/d/1kNHqKI3OL9CeoVJve7ehmWvOSj98G4lQLNubQvTYdpc) I believe the current implementation would allow us to achieve that with the new TURN client with Client#HandleInbound().

## Connectivity check improvement
I have not filed an issue for this, but I guess you may have already realized. This new client can also improve the way we do connectivity checking (with the built-in transaction/retransmission timers).

Sean-Der:
@enobufs this is an A+ for me!

The only thing I would push back on (for now) is sharing sockets between candidate types.

I can build a repro via vnet. But right now since we do STUN/TURN via default route it has a much higher chance of not having timed out requests.

If we do the request on all interfaces we can send out requests that don’t route. This blocks startup time for ~5 seconds. Chromium doesn’t have this issue, they seem to be smart enough to only do it on some interfaces.

So I am for sharing sockets. We just need to be careful about which one outbound STUN/TURN requests happen on I think.

enobufs:
Thanks @Sean-Der , good to know you are confortable with the direction!

Just one thing to make sure we are on the same boat.. (please bare with me)

As you mentioned "default interface", I think you are talking about "multi-homed" cases (where we have more than one local IP address or NICs.  That is an (separate) issue we will need to sort out but what I'd like to address soon is a different thing which is mentioned in pion/ice#74.

Even there is only one local IP (non-loopback) on one NIC, we are allocating three UDP sockets (conns) for each candidate types when gathering candidates.

1. one for initial host candidate [here](https://github.com/pion/ice/blob/v0.4.3/gather.go#L173)
1. one for srflx candidate [here](https://github.com/pion/ice/blob/v0.4.3/gather.go#L243)
1. one for relay candidate [here](https://github.com/pion/ice/blob/v0.4.3/gather.go#L313)

Where we should use the same `conn` allocated for host candidate for srflx and relay candidates.  This causing more pings, NAT bindings and unnecessary prflx candidates (wasteful in resources).

Here is what it looks like:
Screen Shot 2019-07-06 at 4 13 24 PM

client.go
enobufs: Done

internal/client/transaction.go
enobufs: Done

enobufs:
I am testing with coturn.  It returns 401 error to the refresh request.  I'm looking into it.

enobufs:
@Sean-Der This is good to go.  I am pretty confident because it is working with coturn, too! Let's unblock pion/ice#46!

enobufs:
@Sean-Der @hugoArregui 
In case that helps, please take a look at the new TURN client example (`turn-client`) [here](https://github.com/pion/turn/tree/cmd2examples/examples) (in another branch)

I think this demonstrates how the new turn.Client's API looks like and works, with a hope that helps your review!

enobufs:
Also, this is [my implementation note](https://docs.google.com/document/d/1kNHqKI3OL9CeoVJve7ehmWvOSj98G4lQLNubQvTYdpc)

client.go
```diff
@@ -101,3 +217,346 @@ func (c *Client) SendSTUNRequest(serverIP net.IP, serverPort int) (net.Addr, err
...
+
+// HandleInbound handles data received.
+// This method handles incoming packet demultiplex it by the source address
+// and the types of the message.
+// This return a booleen (handled or not) and if there was an error.
+// Caller should check if the packet was handled by this client or not.
+// If not handled, it is assumed that the packet is application data.
+// If an error is returned, the caller should discard the packet regardless.
+func (c *Client) HandleInbound(data []byte, from net.Addr) (bool, error) {
+	var handled bool
```
Sean-Der: I would return values right away. I think this works great right now, but seems really easy to make a mistake in the future when editing.

```
	case stun.IsMessage(data):
		return true, c.handleSTUNMessage(data, from)
	case len(c.turnServStr) != 0 && from.String() == c.turnServStr:
		if turn.IsChannelData(data) {
			return true, c.handleChannelData(data)
		} else {
		       return true,  fmt.Errorf("unexpected packet from TURN server")
		}
```

enobufs: Agreed. I have modified it as suggested.

enobufs:
@Sean-Der any suggestion for the version number? (last ver was v1.2.1)

PR #85, opened 2019-08-15
Allocation fresh bug fixed

### LINKED ISSUE #84: Allocation refresh with zero lifetime not deleted as expected
### Your environment.
  - Version: v1.3.3 -38ac12
  - Browser: n/a

### What did you do?

Refresh allocation with 0 second lifetime

### What did you expect?
The `allocation` doesn't remove from the manager as expected.
### What happened?
Not removed, this because when we use `0` fresh lifetime, the old lifetimer only stop rather firing at once, related code is [allocation.go#L165](https://github.com/pion/turn/blob/master/internal/allocation/allocation.go#L165).

songjiayang:
Fixed #84


enobufs:
I removed ./internal/proto/uri.go because it had a bug to fail CI test but no one is using it.
The bug is, `net/url`'s Parse() works with a url like `turn://example.org:1234` but NOT like `turn:example.org:443` (no `//`). The behavior was inconsistent (e.g. passe on mac, but not on CI, etc).

songjiayang:
@enobufs Thanks

PR #86, opened 2019-08-18
Add external IP mapping

### LINKED ISSUE #56: Support of external IP address mapping

enobufs:
Resolves #56

See #56 for the motivation of this PR.

### DIFF
--- a/internal/allocation/allocation_manager.go
+++ b/internal/allocation/allocation_manager.go
@@ -21,6 +21,7 @@ type ManagerConfig struct {
 type Manager struct {
 	lock        sync.RWMutex
 	allocations map[string]*Allocation
+	extIPMap    map[string]net.IP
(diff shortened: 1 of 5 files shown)


enobufs:
Thanks @songjiayang !

PR #89, opened 2019-08-29
Remove AtomicBool closed from conn

Sean-Der:
Early return instead of setting+checking in a loop


enobufs:
Is this related to the deadlock (on Close()) discussed in the slack channel?

Sean-Der:
@enobufs yea! I have been reading through the TURN code just to see if anything pops out at me.

Looking through `Client` I don't see anywhere that things could deadlock so I think we are A+ in this code :)

We have had a few possibly scary changes lately in `pion/ice` so I am thinking it might be one of those.

PR #93, opened 2019-09-11
Fix nil transaction result message

lrezek:
#### Description
When we're waiting for a transaction result and the channel is closed, `Msg` will be nil within the returned result. Since result.Msg is in use all over the place, this fixes various nil pointer issues without having to go through all the usages and do nil checks.


Sean-Der:
Fantastic thank you @lrezek !

lrezek:
Hey @enobufs, yup, that's definitely better. I see now that the caller of that function does check `result.Err`, so this'll work 👍

Sean-Der:
Thank you so much @enobufs :)

Fantastic work @lrezek! If you are interested we are all on [Slack](https://pion.ly/slack) if you ever want to chat/have questions.

PR #98, opened 2019-09-22
Fix incorrect warning

### LINKED ISSUE #97: Incorrect warning: "vnet is enabled"
### Your environment.
  - Version: v1.3.6
  - Browser: n/a

### What did you do?
Ran my app using pion/turn with PION_LOG_WARN=all

### What happened?

I see this warning even though vnet is not enabled
```
turnc WARNING: 2019/09/21 19:48:15 vnet is enabled
```

enobufs:
Resolves #97


client.go
```diff
@@ -78,7 +78,7 @@ func NewClient(config *ClientConfig) (*Client, error) {
 
 	if config.Net == nil {
 		config.Net = vnet.NewNet(nil) // defaults to native operation
-	} else {
+	} else if config.Net.IsVirtual() {
```
hugoArregui: should we do this for ice as well? (https://github.com/pion/ice/blob/master/agent.go#L345)

enobufs: Yes, I have [this commit](https://github.com/pion/ice/pull/103/commits/844561e47b58c47b1c95c9c5dd90219fb31c56ad)

Issue #102, opened 2019-09-30
CreatePermission error response (error 438: Stale nonce)


enobufs:
Reported by Aaron France
### Your environment.
  - Version: v1.3.7?

The log comes from [this line](https://github.com/pion/turn/blob/541c905657714a0127be801c83dfec09860131ff/internal/client/conn.go#L489)

* TURN connection was not used at that time
* It appears to happen after 15 min or so.

enobufs:
There seems to be two issues:

1. Why server is saying stale noce but no refreshed?
1. Why ICE keeps using TURN connection which was not "selected"

enobufs:
@AeroNotix
Not too informative, Chromium had a similar problem long time ago: https://bugs.chromium.org/p/webrtc/issues/detail?id=1520
```
Comment 5 by juberti@google.com on Wed, Mar 20, 2013, 12:35 PM PDT
Hmm, I forgot to implement this.  Should be a 10 line change or so - if we get a 438, just blow away TurnPort::hash_, since it's now invalid, and retry.

To test, suggest adding a method on TurnServer that expires the nonce after every request, or something like that.
```

AeroNotix:
@enobufs the solution mentioned in that thread is relatively the same as what I have in the PR _except_ they _seem_ to be suggesting what I alluded to in: https://github.com/pion/turn/pull/103#issuecomment-536927031

Meaning anytime we get a 438 from any request, we need to fallback, get the new nonce and retry. I'd actually recommend proactively getting a new nonce way before any of the agreed upon timers are done.

PR #103, opened 2019-09-30
fix: update nonce in refreshPermissions when refresh is 438

AeroNotix:
Relates: #102 

cc @enobufs

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,7 @@ Check out the [CONTRIBUTING.md](CONTRIBUTING.md) to join the group of amazing pe
 * [Robert Eperjesi](https://github.com/epes)
 * [Lukas Rezek](https://github.com/lrezek)
 * [Hugo Arregui](https://github.com/hugoArregui)
+* [Aaron France](https://github.com/AeroNotix)
 
 ### License
 MIT License - see [LICENSE.md](LICENSE.md) for full text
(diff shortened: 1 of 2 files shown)


AeroNotix:
Build failed because of `The command "bash .github/lint-commit-message.sh" exited with 1.` All worthwhile tests pass, however.

Sean-Der:
@AeroNotix thanks for grabbing this!

CI should be green if
* Add yourself to the README
* Keep your subject to 50 chars `fix: handle stale nonce in refreshPermissions`  you can fill out the body more, could be helpful for others in the future.

AeroNotix:
@enobufs yes this does fix the issue for me.

Here's how I tested:

* Two applications, one with and one without the code in the PR. 
* I brokered a connection between the two peers and waited 12 minutes.
    * 12 minutes because _by default_ (per RFC) the nonce value is valid for 10 minutes, we then have a timer on 120 seconds which calls CreatePermissions. To me I believe this makes the timing of the 438 line up to when the server has invalidated the nonce value.
* Once the 12 minutes was up, I ensured to check logs. In the peer without the change, the CreatePermissions 438 error happened regularly in the logs.
* Once the 12 minutes was up, in the peer _with_ the changes the logs were free from the CreatePermissions 438 error.

I believe, for my _particular_ use-case this PR fixes the issue but reading through the code more and more and starting to understand the RFC for how the Nonce should/could be used I believe a more general way to update the nonce could be implemented.

enobufs:
(I really don't like this commit-message checker...)
`git rebase master -i` should help you modifying the commit message...

```
* Capitalize the subject line
```

AeroNotix:
@enobufs now the linter is happy now

enobufs:
@AeroNotix Great work! Thank you!

PR #104, opened 2019-10-06
Add STUNConn. Wrapping net.Conn for TURN Clients

Sean-Der:
This allow us to use a net.Conn (like TCPConn) as the conn
for a TURN client.

Relates to #1

### DIFF
--- a/internal/proto/chandata.go
+++ b/internal/proto/chandata.go
@@ -10,11 +10,10 @@ import (
 //
 // See RFC 5766 Section 11.4
 type ChannelData struct {
-	Data    []byte // can be subslice of Raw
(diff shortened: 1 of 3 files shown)


internal/proto/chandata.go
```diff
@@ -58,9 +57,6 @@ func (c *ChannelData) Encode() {
 	c.Raw = c.Raw[:0]
 	c.WriteHeader()
 	c.Raw = append(c.Raw, c.Data...)
-	if !c.Padding {
```
Sean-Der: Since we don't have a way to distinguish `Stream` vs `Datagram` transport this is easier for now.

We could make this a bool, but I am not sure the complexity is worth it. Happy to re-do if you disagree :)

PR #112, opened 2020-01-13
Update CI configs to v0.1.0

pionbot:
Update lint scripts and CI configs.

PR #114, opened 2020-01-13
Fix Go 1.12 build

at-wat:
Error wrapping is not yet supported on Go 1.12 fmt package.
It would be nice to be migrated to "%w" error wrapping once
Go 1.12 is deprecated.

ref: https://github.com/pion/turn/pull/112

@Sean-Der any ideas on it?


Sean-Der:
@at-wat works for me! 

I need to write up a nice script to find all the times we pass `err` or `error` to a `fmt.Printf` so we can gradually move over to `%w` when we are ready :)

Thanks for catching this!

PR #124, opened 2020-02-07
Fix "no listener for" race condition on fast server response

tclift:
#### Description

Fix race condition where we could attempt to process a result before an appropriate handler was in place. This resulted in log lines like:
```
turnc DEBUG: 23:36:54.500219 client.go:489: no listener for Allocate error response l=88 attrs=5 id=WlJEHi6/lMtgixTn
```
This happened frequently when two peers and the STUN/TURN server were on the same host (e.g. during development, running tests, performing a server-local health check).

The issue was when sending on a transaction's unbuffered result channel before a goroutine was started to receive on the channel. By adding a buffer size of one, it doesn't matter if the result is sent slightly earlier, and we can avoid the complexities of handling concurrency with an unbuffered channel.

There is a related concurrency issue here. A caller of `Transaction.WaitForResult` (or `Client.PerformTransaction`) will block until the transaction's internal `resultCh` channel is closed. This channel is only closed if `Transaction.Close` is called. `Transaction.Close` is only called from `TransactionMap.CloseAndDeleteAll`, which is only called when the client is closed. The transaction is not closed e.g. on initial write failure or retransmission limit exceeded. Should we consider calling `Transaction.Close` in these situations? Care would need to be taken to avoid writing to the closed channel. Aside from the `WaitForResult` blocking, this commit is adding more memory (the now buffered result) that would not be collected until the transaction itself is collected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,7 @@ Check out the [CONTRIBUTING.md](CONTRIBUTING.md) to join the group of amazing pe
 * [Hugo Arregui](https://github.com/hugoArregui)
 * [Aaron France](https://github.com/AeroNotix)
 * [Atsushi Watanabe](https://github.com/at-wat)
+* [Tom Clift](https://github.com/tclift)
 
 ### License
 MIT License - see [LICENSE.md](LICENSE.md) for full text
--- a/client.go
+++ b/client.go
@@ -335,17 +335,19 @@ func (c *Client) Allocate() (net.PacketConn, error) {
 }
 
 // PerformTransaction performs STUN transaction
-func (c *Client) PerformTransaction(msg *stun.Message, to net.Addr, dontWait bool) (client.TransactionResult, error) {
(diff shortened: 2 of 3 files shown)


tclift:
Updated implementation to perform a blocking send on the (unbuffered) result channel, rather than a non-blocking send on a buffered channel.

tclift:
Oops, forgot to re-run that after updating the msg. Done now.

Sean-Der:
Fantastic work @tclift merged!

* Would you like me to tag a new version?
* Would you like to join the Pion github org? You can then review PRs, and merge your own PRs when they are reviewed!

No pressure though. If you find WebRTC interested and are looking to be involved in Open Source we would love to have you :) Feel free to aggressively change/refactor TURN as well, as long as the tests pass and the public API is stable everything else is fair game.

tclift:
I see you tagged a release, thanks for that.

Thanks very much for the offer to join the org. I won't take you up on it just yet... I'll see how deep I need to get as my project progresses. I might be looking to add ORIGIN support so that authentication realm can be used in coturn.

PR #153, opened 2020-08-24
Reactualized examples because of V2

iashchak:
#### Description
Added go.mod files for server examples
Updated shell commands to build project

#### Reference issue
n/a

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -20,16 +20,31 @@ This example logs all inbound/outbound STUN packets. This could be useful if you
 
 You could also intercept these reads/writes if you want to filter traffic going to/from specific peers.
 
+```sh
+$ cd log
+$ go mod download
+$ go build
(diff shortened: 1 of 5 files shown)


Sean-Der:
Thanks @iashchak!

Thanks for updating the docs. That is a great change.

I don't think we want to add the `go.mod` files. Right now if you `go get` it will use the latest version of pion/turn (or if doing `go build` will use the code you currently have)

If we do `go.mod` files then we have to remember to bump them. If Go projects are supposed to have mod files for examples we can do though that!

iashchak:
Hello @Sean-Der !
Thank you for your review.
I am a kind of newbie in the go-lang world, so i just found that its impossible to build it using the doc until you have library pre-installed.
Mod file is a just one of the ways to build examples. Another - is an adding package via 
```sh 
go get
```
If its useless - feel free to close the PR :)

Anyway, have a great day!

PR #164, opened 2020-09-17
Long-Term Credentials (lt-cred mechanism)

hermanbanken:
#### Description
Context: For Jitsi I was first using coTURN and then stumbled upon Pion. What I liked was owning the server code so that I could diagnose authentications. I was having trouble with that with coTURN I believed.

Jitsi proposes to use ephemeral credentials. They reference a draft from their [TURN documentation](https://jitsi.github.io/handbook/docs/devops-guide/turn). The idea is to use the expiry time of the credential as the username, and let the password contain some cryptographic hash of a (server-side) shared-secret and the expiry time.

This is the same mechanism used by coTURN and it is documented here how to generate a quick pair of credentials if you have the shared-secret:
https://meetrix.io/blog/webrtc/coturn/installation.html#testing

# RFC
Long-Term Credential Mechanism (RFC5389-10.2: https://tools.ietf.org/search/rfc5389#section-10.2)

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ Check out the [CONTRIBUTING.md](CONTRIBUTING.md) to join the group of amazing pe
 * [lllf](https://github.com/LittleLightLittleFire)
 * nindolabs (Marouane)
 * [Onwuka Gideon](https://github.com/dongido001)
+* [Herman Banken](https://github.com/hermanbanken)
 
 ### License
 MIT License - see [LICENSE.md](LICENSE.md) for full text
--- a/examples/turn-client/lt-cred/README.md
+++ b/examples/turn-client/lt-cred/README.md
@@ -1,0 +1,16 @@
+# Usage
+
+```bash
+export SECRET=somesecret
+
+# Build binaries
+(cd examples/turn-client/lt-cred && go build .)
+(cd examples/turn-server/lt-cred && go build .)
+(cd examples/turn-client/tcp && go build .)
+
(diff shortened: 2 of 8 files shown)


Sean-Der:
This is amazing! Great work @hermanbanken 

Thanks for sharing, it would have been super easy for you not to Open Source this :) 

Would you mind adding some basic test coverage? If not I can grab it this weekend!

hermanbanken:
Doesn't the missing test coverage come from the rest of the code somehow? I don't really understand the gigantic drop in coverage.

Or do examples also count towards the coverage?

I don't have the time this Friday afternoon (7 PM in CEST) or the weekend, so please go ahead!

Sean-Der:
@hermanbanken oh yea I think you are right! I will re-run all the builders and I bet will go green :)

thanks for the contribution, I got it from here. Just our messed up infra -_-

at-wat:
@Sean-Der It seems that the test configuration got wrong since https://github.com/pion/turn/pull/161#issuecomment-689361448. (-23% coverage)

Sean-Der:
Merged! Great work @hermanbanken 

I made a few changes from your original PR

* Update `tcp` and `udp` client to accept passwords with `=` in them.
* Move ltcred out of `internal`. No one would be able to import that code out of tree.
* `NewLongTermAuthHandler` takes a `logging.LeveledLogger` now.

hermanbanken:
Thanks!

PR #172, opened 2020-10-11
Add port range relay generator

### LINKED ISSUE #129: port range
port range feature will be so nice if consider server is behand firwall. Do you will consider add this fearture? thank you
Ps the project is so good. By mixing variable technology tools, golang can finnaly bulid an webrtc engine, and mybe do extruct some ideas to build another net streamming commication structure.

jmattheis:
This PR adds a relay generator with a port range restriction.

Fixes #129


relay_address_generator_range.go
```diff
@@ -0,0 +1,105 @@
...
+		return errRelayAddressInvalid
+	case r.Address == "":
+		return errListeningAddressInvalid
+	default:
+		return nil
+	}
+}
+
+func (r *RelayAddressGeneratorPortRange) randPort() uint16 {
+	r.randMutex.Lock()
```
jmattheis, line 67: I don't know if AllocatePacketConn can be called concurrently, thus, I preemptively used a mutex to lock the not threadsafe random source.

relay_address_generator_range.go
```diff
@@ -0,0 +1,105 @@
...
+	defer r.randMutex.Unlock()
+	return r.MinPort + uint16(r.Rand.Intn(int((r.MaxPort+1)-r.MinPort)))
+}
+
+// AllocatePacketConn generates a new PacketConn to receive traffic on and the IP/Port to populate the allocation response with
+func (r *RelayAddressGeneratorPortRange) AllocatePacketConn(network string, requestedPort int) (net.PacketConn, net.Addr, error) {
+	if requestedPort != 0 {
```
jmattheis, line 74: Is this needed? I've looked through the code and could only find requested port 0.

relay_address_generator_range.go
```diff
@@ -0,0 +1,105 @@
...
+		relayAddr := conn.LocalAddr().(*net.UDPAddr)
+		relayAddr.IP = r.RelayAddress
+		return conn, relayAddr, nil
+	}
+
+	for try := 0; try < r.MaxRetries; try++ {
+		conn, err := r.Net.ListenPacket(network, fmt.Sprintf("%s:%d", r.Address, r.randPort()))
+
+		if err != nil {
+			if strings.Contains(err.Error(), "address already in use") {
```
jmattheis, line 88: Is this okay? it seems a little hacky to me but I couldn't find a better solution.

jmattheis (commented):
Questions in subcomments.

relay_address_generator_range.go
```diff
@@ -0,0 +1,104 @@
...
+}
+
+// Validate is called on server startup and confirms the RelayAddressGenerator is properly configured
+func (r *RelayAddressGeneratorPortRange) Validate() error {
+	if r.Net == nil {
+		r.Net = vnet.NewNet(nil)
+	}
+
+	if r.Rand == nil {
+		r.Rand = rand.New(rand.NewSource(time.Now().UnixNano()))
```
jmattheis, line 45: Can I ignore gosec. 
> Use of weak random number generator (math/rand instead of crypto/rand) (gosec)

Using a crypto rand source isn't really critical here?

Sean-Der:
Nice! I was so excited to see you using Pion TURN @jmattheis screego is really great. Would you be ok if I added it to awesome-pion ?

relay_address_generator_range.go
Sean-Der: I would check out https://github.com/pion/randutil it doesn't do much, just cuts down on the amount of duplication we need to do.

relay_address_generator_range.go
Sean-Der: This is used for `EVEN-PORT` and `RESERVATIONS`. This is never used in WebRTC TURN clients though.

I think this check makes sense!

relay_address_generator_range.go
Sean-Der: I would drop the `string.Contains`. Might be worth logging the error though!

relay_address_generator_range.go
Sean-Der: LGTM! With randutil you might be able to drop this entirely.

relay_address_generator_range.go
jmattheis: But we need the check (or something similar). Otherwise, AllocatePacketConn will err and not retry if the rand source generates an already binded port.

relay_address_generator_range.go
jmattheis: Nice, didn't knew this existed (:.

jmattheis:
I'll do it tomorrow, in my timezone it is already pretty late. Yeah, it would be ok to add screego to this awesome list. 

pion/turn is great, thank you really much for it! Originally I wanted to implement screego in rust, but the existing turn library for rust wasn't as good as this one. So pion/turn is responsible that screego was written Go (:. 

I actually had no problems integrating pion/turn, it has a simple api for integration and most of the methods/parameters are self-explanatory. Maybe it would be good, to have some kind of explanation of how turn is implemented on a surface level. F.ex. that it does allocate udp ports.

relay_address_generator_range.go
Sean-Der: Sorry I rushed my comment :( I would do this. Don't worry about catching the specific error, since it isn't typed I could see this causing issues on platforms we don't test for.

```
if err != nil {
  continue
}
```

Sean-Der: If you agree with that I would love to merge (and tag) this today! Happy to make the change myself and squash + push (with you as the author!)

jmattheis: Yeah, this is fine by me :+1:.

Sean-Der:
Merged with 73942c387bd12c0d200bb0c6813421cb352cf77c

Thank you so much for doing this @jmattheis I also tagged `v2.0.5`

PR #193, opened 2020-12-28
Fix #192 - Configurable server inboundMTU

### LINKED ISSUE #192: turn ERROR: BadFormat for attribute/message: buffer length 1500 is less than 1520 (expected message size)
### Your environment.
  - Version: *feb51eca38f7f449bee345723b26b44202b5e590*
  - Browser: *include version*

### What did you do?

Currently building an SFU. While testing I've set (sdpSemantics: 'unified-plan', iceTransportPolicy: 'relay') to force the traffic to be sent via turn server.
When adding 2 or more streams to a peer connection it seems the payload gets too big causing turn errors.

### What did you expect?

No errors should be shown.

### What happened?

turn ERROR: 2020/12/12 17:22:52 error when handling datagram: failed to create stun message from packet: BadFormat for attribute/message: buffer length 1500 is less than 1520 (expected message size)

andrefsp:
#### Description
This PR creates a server option so the inbound MTU may be configured. Currently hardcoded to 1500 bytes it may sometimes cause issues when the stun/turn payload becomes to big. In such cases it seems that increasing max payload size fixes the issue.

#### Reference issue
Fixes #192

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,8 @@ Check out the [CONTRIBUTING.md](CONTRIBUTING.md) to join the group of amazing pe
 * [Onwuka Gideon](https://github.com/dongido001)
 * [Herman Banken](https://github.com/hermanbanken)
 * [Jannis Mattheis](https://github.com/jmattheis)
+* [andre da palma](https://github.com/andrefsp)
+
 
 ### License
 MIT License - see [LICENSE.md](LICENSE.md) for full text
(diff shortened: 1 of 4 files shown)


Sean-Der:
Merged with 13867664acbcf7a2b55f561fc4ed61b46638438d Thank you so much for doing this @andrefsp 

Would you like being part of the Pion org? Would love to have you help drive the project. Also if you have any important features would love to work on them together. Anything I can do to make this project healthy/get users involved :)

PR #205, opened 2021-02-28
Avoid infinite loop in GetRandomEvenPort

jech:


Sean-Der:
Thanks @jech !

Would you like me to tag a new version?

Sean-Der (approved):

jech:
> Would you like me to tag a new version?

No need, it's just a minor cleanup that i did while I get to grok the TURN code.  I don't think the infinite loop can actually trigger on a Linux kernel.

PR #232, opened 2021-09-08
Close accepted TURN listener connections

deansheather:
#### Description
To the best of my knowledge, connections accepted from the provided net.Listener would never be closed, even if there was an error.

This changes them to be closed at the top of the connection handling goroutine. I didn't touch the connections from packet conn configs because I didn't want to change the behavior (parent code might already be doing `defer conn.Close()`).

I had to refactor the NewServer method a bit to avoid hitting the gocognit linter. The only real changes I made were to the connection handler goroutine in the net.Listener goroutine.

#### Reference issue
None

cc @coadler @kylecarbs

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -8,6 +8,7 @@ Aleksandr Razumov <ar@gortc.io>
 andrefsp <andrefsp@gmail.com>
 Atsushi Watanabe <atsushi.w@ieee.org>
 backkem <mail@backkem.me>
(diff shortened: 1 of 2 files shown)


stv0g (changes requested):
Unfortunately, there are some conflicts with the recent master branch.

Can you rebase your PR and fix the conflicts?

stv0g:
I created a new PR based on the latest master in #306  as this seems to be outdated.

PR #233, opened 2021-11-03
Allocation count methods

frostbyte73:
#### Description

Adds an allocation count method to the server, returning the number of active allocations. This is useful for draining, where you can stop routing traffic to an instance and then wait until `server.AllocationCount() == 0` before shutting down.


davidzhao (approved):
lg!

PR #242, opened 2022-02-12
Add CreatePermission method to the TURN client

ted-skydio:
#### Description
This adds a `CreatePermission()` method to the turn client, which sends a CreatePermission request as described in [RFC 5766, Section 9](https://datatracker.ietf.org/doc/html/rfc5766#section-9).

This method allows a TURN client to request or refresh permissions for a specific set of IP address / port pairs.

* Makes the existing `UPDConn.CreatePermission()` method public.
* Adds a `CreatePermission()` method on `Client` that exposes the underlying `relayedUDPConn().CreatePermissions()` method.

#### Reference issue
There's not a github issue for it. I just needed this functionality and discovered it wasn't available and couldn't be achieved with the library as-is due to the `CreatePermissionRequest` and `PeerAddress` types being in an internal package.

### DIFF
--- a/client.go
+++ b/client.go
@@ -333,6 +333,12 @@ func (c *Client) Allocate() (net.PacketConn, error) {
 	return relayedConn, nil
 }
 
+// CreatePermission Issues a CreatePermission request for the supplied addresses
+// as described in https://datatracker.ietf.org/doc/html/rfc5766#section-9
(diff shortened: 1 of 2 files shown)


Sean-Der (approved):

Sean-Der:
Thanks @ted-skydio ! I also tagged `v2.0.7`

PR #252, opened 2022-04-08
lt_cred support for TURN REST format

giavac:
#### Description

This adds support for the format timestamp:username used with the TURN REST API (https://tools.ietf.org/search/rfc5389#section-10.2).

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -9,6 +9,7 @@ andrefsp <andrefsp@gmail.com>
 Antonio Sorrentino <ansorrenltd@gmail.com>
 Atsushi Watanabe <atsushi.w@ieee.org>
 backkem <mail@backkem.me>
+Caleb Phillips <calebrphillips@gmail.com>
 cnderrauber <zengjie9004@gmail.com>
 David Colburn <xero73@gmail.com>
 Herman Banken <hermanbanken@gmail.com>
--- a/FAQ.md
+++ b/FAQ.md
@@ -1,0 +1,17 @@
+
+
+## FAQ
+
+Q: Will pion/turn also act as a STUN server?
+
+A: Yes.
+
+Q: How do I implement token-based authentication?
+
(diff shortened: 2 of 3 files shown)


stv0g (approved):
Thanks @giavac!

This is actually really useful for me as well :)
Thanks for the contribution!

guusdk:
Hi! We're experiencing a challenge with Pion using REST API compatible (ephemeral) authentication. This may well be in our implementation, but just to make sure we're not chasing ghosts: Has the change mentioned in this PR made it into any release to date? The commit references are a bit tricky to follow through the Github UI.

giavac:
My changes were removed with the force-push mentioned in this PR. I'm not sure what happened, I might be missing something from the github flow.
I had a `GenerateLonTermTURNRESTCredentials()` and `LongTermTURNRESTAuthHandler()` functions to handle the long term credentials support, with an example.
If this is still needed I can try and create a new PR?
Otherwise you can retrieve the missing code from the diff of the forced push.

guusdk:
The functionality is still desirable. As I'm not familiar with the code base (or even the programming language), I'd prefer you (instead of me) recreating the commit.

rg0now:
I agree: this is still a useful feature to have. We have [implemented this functionality in STUNner](https://github.com/l7mp/stunner/blob/089b1a9c14ef26e098c17bbb522c6397f4ff0549/handlers.go#L37), with the twist that we consider the first thing that parses as a valid unsigned integer as a UNIX timestamp, so we accept both `timestamp:username` and `username:timestamp`, and also a plain `timestamp` (without a user name) or even `x:y:z:timestamp` as well. This is completely upward compatible with the handler generated by the default [`NewLongTermAuthHandler`](https://github.com/pion/turn/blob/0c34bf2ca7254796cabbe086fac087836d0c92b9/lt_cred.go#L37), so you don't even need to introduce a separate `LongTermTURNRESTAuthHandler`. 

Please submit as a new PR, I'd be happy to review.

giavac:
Sounds good; it's been more than a year not using this code base so I'll need a little more time than just recreating the PR (also based on @rg0now 's information).

giavac:
I've raised https://github.com/pion/turn/pull/370 - CI checks are pending but for the people interested in this feature please take a look and review.

PR #265, opened 2022-06-04
Add IPv6 ClientConfig flag

hetelek:
Adds a `IPv6` flag to the turn.ClientConfig, which will allow for IPv6 relay support.

https://github.com/pion/ice/pull/462

### DIFF
--- a/client.go
+++ b/client.go
@@ -43,6 +43,7 @@ type ClientConfig struct {
 	Conn           net.PacketConn // Listening socket (net.PacketConn)
 	LoggerFactory  logging.LoggerFactory
 	Net            *vnet.Net
+	IPv6           bool
 }
 
 // Client is a STUN server client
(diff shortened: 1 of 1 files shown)


Sean-Der:
@jech

jech:
@hetelek could you please explain why this is needed?  Why is it not enough to infer the address family from ClientConfig.Conn, using something like:

   nework := config.Conn.LocalAddr().Network()

This would avoid adding a new API, and would give us more flexibility in the future if we decide that this is not the right approach.

If an explicit field is needed, I'd prefer it to be called `Network` and to be either a string or an int rather than a simple boolean, in case more values are needed in the future.

stv0g:
This PR gets obsoleted if #276 gets merged.

stv0g:
#276 has been merged. I therefore close this PR.

Issue #134, opened 2020-04-07
Add PermissionHandler to ServerConfig


Sean-Der:
Right now users have no way to filter/log what permissions are being created. An attacker could be using the TURN server as a bastion box into NAT.

https://www.rtcsec.com/2020/04/01-slack-webrtc-turn-compromise/

stv0g:
Solved by #267

PR #267, opened 2022-06-09
Add PermissionHandler to ServerConfig #134 

rg0now:
#### This is another attempt to address #134, see an earlier attempt in #222

#222 introduces the DeniedPeerRange stanza into the ServerConfig to implement peer address blacklisting. This approach has a couple of issues:
- implements only peer blacklists, but does not allow whiletelisting or filtering based on DNS, etc.
- handles only the ChannelBindRequest codepath, but leaves the CreatePermission codepath (https://datatracker.ietf.org/doc/html/rfc8656#section-3.4) open 
- introduces a new package dependency on "inet.af/netaddr"

This patch takes a different approach: it allows the user to specify a PermissionHandler callback for each PacketConnConfig/ListenerConfig in the ServerConfig. Whenever a permission is about to be created via the associated PacketConn/Listener (either via a ChannelBindRequest or a CreatePermission), the PermissionHandler is called with the requested peer address and it can decide whether to accommodate the permission request (return boolean true) or deny it (return false). In the latter case, a "permission request administratively prohibited" error is returned to the client.

Also added tests and an example.

### DIFF
--- a/internal/server/turn.go
+++ b/internal/server/turn.go
@@ -231,8 +231,15 @@ func handleCreatePermissionRequest(r Request, m *stun.Message) error {
 			return err
 		}
 
+		if err := r.AllocationManager.GrantPermission(r.SrcAddr, peerAddress.IP); err != nil {
+			r.Log.Infof("permission denied for client %s to peer %s", r.SrcAddr.String(),
+				peerAddress.IP.String())
+			return err
+		}
+
 		r.Log.Debugf("adding permission for %s", fmt.Sprintf("%s:%d",
 			peerAddress.IP.String(), peerAddress.Port))
+
 		a.AddPermission(allocation.NewPermission(
 			&net.UDPAddr{
 				IP:   peerAddress.IP,
@@ -314,6 +321,16 @@ func handleChannelBindRequest(r Request, m *stun.Message) error {
 	peerAddr := proto.PeerAddress{}
 	if err = peerAddr.GetFrom(m); err != nil {
 		return buildAndSendErr(r.Conn, r.SrcAddr, err, badRequestMsg...)
+	}
+
+	if err = r.AllocationManager.GrantPermission(r.SrcAddr, peerAddr.IP); err != nil {
+		r.Log.Infof("permission denied for client %s to peer %s", r.SrcAddr.String(),
+			peerAddr.IP.String())
+
+		unauthorizedRequestMsg := buildMsg(m.TransactionID,
+			stun.NewType(stun.MethodChannelBind, stun.ClassErrorResponse),
+			&stun.ErrorCodeAttribute{Code: stun.CodeUnauthorized})
+		return buildAndSendErr(r.Conn, r.SrcAddr, err, unauthorizedRequestMsg...)
 	}
 
 	r.Log.Debugf("binding channel %d to %s",
--- a/server_test.go
+++ b/server_test.go
@@ -114,6 +114,156 @@ func TestServer(t *testing.T) {
 		assert.Equal(t, server.inboundMTU, 2000)
 		assert.NoError(t, server.Close())
 	})
+	t.Run("Filter on client address and peer IP", func(t *testing.T) {
+		udpListener, err := net.ListenPacket("udp4", "0.0.0.0:3478")
+		assert.NoError(t, err)
+
+		server, err := NewServer(ServerConfig{
+			AuthHandler: func(username, realm string, srcAddr net.Addr) (key []byte, ok bool) {
+				if pw, ok := credMap[username]; ok {
+					return pw, true
+				}
+				return nil, false
+			},
+			PacketConnConfigs: []PacketConnConfig{
+				{
+					PacketConn: udpListener,
+					RelayAddressGenerator: &RelayAddressGeneratorStatic{
+						RelayAddress: net.ParseIP("127.0.0.1"),
+						Address:      "0.0.0.0",
+					},
+					PermissionHandler: func(src net.Addr, peer net.IP) bool {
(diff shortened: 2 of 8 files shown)


yawn:
This looks like a great contribution! Anything anyone outside of the PR can do to help move it along?

server_test.go
```diff
@@ -23,6 +23,7 @@ func TestServer(t *testing.T) {
 	defer report()
 
 	loggerFactory := logging.NewDefaultLoggerFactory()
+	// loggerFactory.DefaultLogLevel = logging.LogLevelTrace
```
stv0g, line 26: Can we remove the dead code?

internal/server/turn.go
```diff
@@ -76,7 +76,7 @@ func handleAllocateRequest(r Request, m *stun.Message) error {
 	// 5.  The server checks if the request contains a RESERVATION-TOKEN
 	//     attribute.  If yes, and the request also contains an EVEN-PORT
 	//     attribute, then the server rejects the request with a 400 (Bad
-	//     Request) error.  Otherwise, it checks to see if the token is
```
stv0g, line 79: Seems like an odd change? Why is there a tab?

stv0g (approved):
I did a review of the PR. To me it looks good overall.

From my side, we could merge it :)

server_test.go
rg0now: Thanks! Fixed per [f3af4ad](https://github.com/pion/turn/pull/267/commits/f3af4ad1a272d8a2a46a8170b983a3641e6d1dfc).

internal/server/turn.go
rg0now: Fixed per [c5dca04](https://github.com/pion/turn/pull/267/commits/c5dca04e772ad98d54da1409afc8803b2d7900de).

stv0g:
Thanks @rg0now for addressing my comments.

To get this merged, we need to rebase it on the latest master and fix the commit messages as they cause the CI to fail.

Do you have the capacity to do it? Or do you need some support?

stv0g:
Thanks @rg0now,

The CI is running again. But the commit messages are still not conformant to the Pion CI checks:

```
The preceding commit message is invalid
it failed 'Separate subject from body with a blank line' of the following checks

* Capitalize the subject line
```

rg0now:
This really baffles me: the corresponding commit message is only a single line (`Fix linter warnings`). Any ideas? Happy to redo the commit if you know how to work around this warning. Plus we fail both tests, with another set of curious errors:
```
Error: Codecov: Failed to properly upload: The process '/home/runner/work/_actions/codecov/codecov-action/v3/dist/codecov' failed with exit code 255
```

stv0g:
I think the Codecov error is only transient.

> the corresponding commit message is only a single line (Fix linter warnings). Any ideas?

I agree. I think the check is also too strict. Usually, most people just repeat the first line separated by an empty line.
But this might also be a sign, that the commits are too fine grained. Maybe we could combine all these commits in a single one with a more detailed description in its message?

stv0g:
There is also another check which fails `golangci-lint`:

>  Error: cognitive complexity 33 of func `NewServer` is high (> 30) (gocognit)

rg0now:
> But this might also be a sign, that the commits are too fine grained. Maybe we could combine all these commits in a single one with a more detailed description in its message?

Ohh, I thought you were going to just squash-merge the whole thing in a single commit. I redid the last commit as per your advice. If this still doesn't work I can go back and re-edit the entire commit history in another branch but I'm afraid that will take another PR. Wdyt?

> Error: cognitive complexity 33 of func NewServer is high (> 30) (gocognit)

I tend to agree with this: `NewServer` was already rather complex in the first place. We've added
just 6 new lines of new fairly trivial code to handle the permissionHandlers.

One way to go about this is to separate out the two loop bodies into a separate function each. Namely, the code that sets up PacketConnConfigs and calls the readLoop from https://github.com/pion/turn/blob/371fc35851d91a885a0621b295f8e9a3ed6af2d2/server.go#L68 to https://github.com/pion/turn/blob/371fc35851d91a885a0621b295f8e9a3ed6af2d2/server.go#L85 could go into a new function, say, `runPacketConn`, and the code that sets up the ListenerConfigs from https://github.com/pion/turn/blob/371fc35851d91a885a0621b295f8e9a3ed6af2d2/server.go#L90 to https://github.com/pion/turn/blob/371fc35851d91a885a0621b295f8e9a3ed6af2d2/server.go#L114 could be refactored into a function called `runListener` or similar. Would you be willing to take such a patch?

rg0now:
On a related note: the fact that there is only a single readLoop per each net.PacketConn limits the max performance of UDP TURN listeners quite substantially, since there is a single thread that handles *all* clients connecting via the listener (see https://github.com/l7mp/stunner/issues/60). This is in contrast to net.Conns, which spawn per-client-connection threads. Would you consider a somewhat intrusive patch to remove this bottleneck?

stv0g:
I agree with both of your comments. However, lets try to get the `PermissionHandler` merged first.

Could you add `//nolint:gocognit` comment above the `NewServer()` function to silence the linter?

Also please squash the commits into a single one.

Afterwards, we can address the two issues which you raised.

rg0now:
> Could you add //nolint:gocognit comment above the NewServer() function to silence the linter?

Done. Fixed the linter warning on my side.

> Also please squash the commits into a single one.

Done.

> Afterwards, we can address the two issues which you raised.

Sure, Looking forward to hearing your opinion.

stv0g:
Thanks a lot @rg0now  🥳  I merged your changes and created two new issues for the problems which you identified.
I am happy to review any PRs closing these issues :)

PR #274, opened 2022-11-15
Fix golangci-lint warnings

stv0g:


ashellunts:
Hi @stv0g, these checks are not enabled in CI, right? Why do you want to fix them?

stv0g:
@ashellunts thats correct. I stumbled upon these warnings as I also run golangci-lint in my editor.

stv0g:
I realized too late that they are not checked by the CI.

So we can also discard this PR..

stv0g:
@ashellunts What do you think. Shall we merge or discard this PR?

ashellunts:
Let's merge it if you want.

ashellunts (approved):

Issue #287, opened 2023-01-18
Multi-threaded readLoop() for UDP TURN Listeners


stv0g:
@rg0now has proposed the following in #267:

> The fact that there is only a single `readLoop()` per each `net.PacketConn` limits the max performance of UDP TURN listeners quite substantially, since there is a single thread that handles all clients connecting via the listener (see https://github.com/l7mp/stunner/issues/60). This is in contrast to net.Conns, which spawn per-client-connection threads. Would you consider a somewhat intrusive patch to remove this bottleneck?

stv0g:
@rg0now has provided a nice example in #295

PR #288, opened 2023-01-19
Implementation of per-client UDP readloops

rg0now:
This is a first attempt to address #287.

The aim is to allow the server to maintain a specific per-client UDP readloop thread instead of one global readloop, which would allow it to drain each client connection on a separate CPU thread. See more on this [here](https://github.com/l7mp/stunner/issues/60).

**Problem:** Currently pion/turn server-side performance is limited at about 40-50 kpps per TURN/UDP listener. This is because we allocate a single *global* `net.PacketConn` per UDP listener, which is then [drained by a single CPU go-routine](https://github.com/l7mp/turn/blob/7bd80d5f800480042e404d1d49e5a6d20377127e/server.go#L68). This means that all client allocations made via that listener will share the same CPU thread and there is no way to load-balance client allocations across CPUs. This only affects TURN/UDP: for TCP, TLS and DTLS the TURN sockets are connected back to the client and therefore [a separate CPU go-routine is created for each allocation](https://github.com/l7mp/turn/blob/7bd80d5f800480042e404d1d49e5a6d20377127e/server.go#L96).

**Proposed solution:** Create a separate `net.PacketConn` per each UDP client-allocation, by (1) sharing the same listener server socket using `SO_REUSEADDR`, (2) connecting each per-allocation connection back to the client, and (3) firing up a separate read-loop/go-routine per each client socket.

**Security:** Care must be taken in implementing this plan: if we blindly create a new socket per received UDP packet then a simple UDP portscan will DoS the TURN listener. In the proposed solution the per-client socket is created only after a client has made a successful allocation, which rules out the blind port-scan problem since the attacker must at least have  a set of valid TURN credentials. It is still not completely safe against DoS attacks though, in that an attacker in possession of valid credentials can still open thousands of server sockets, but this is at least the same level of security as provided by the TURN/TCP implementation.

**Implementation:** Some fairly intrusive changes are required to support this. The patch set in particular:

- moves readLoop() into internal/server/readLoop.go,
- splits server.Request struct into server.State (stuff needed by the readLoop) and server.Request (the rest) that embeds server.State,
- lets the server invoke the Connect callback once it successfully created an TURN allocation for a client to obtain a per-client socket and spawn the readloop go-routine for the new socket,
- adds a WrapConn implementation that wraps the net.Conn returned by the DiapUDP as a net.PacketConn,
- adds an example ( examples/turn-server/udp-connect/main.go) to demonstrate the use of the new feature.

**Performance:** Attached is a super-simple script to benchmark a naive TURN server with (`udp-connect`) and without (`simple`) the patch. Requires the `turncat` TURN client (you can obtain from [here](https://github.com/l7mp/stunner/tree/main/cmd/turncat)) in the source dir, plus iperfv2 in the PATH. The script opens 4 client connections and runs a UDP iperf benchmark in each, then in every 5 seconds it reports the cumulative packet rate through the tested server.  On a 24-core Intel server, the built-in `simple` server (`./test.sh simple 30000000`) produced 34816 pps packet rate and used 141.1% CPU (recall, there are 4 clients connected to the same listener so theoretically the server could scale out to 4 CPUs), while the server with the patches (`./test.sh udp-connect 30000000`) produced 140194 pps with 472% CPU usage.

Feedback appreciated. 

[test.sh.gz](https://github.com/pion/turn/files/10458750/test.sh.gz)

### DIFF
--- a/examples/turn-server/udp-connect/main.go
+++ b/examples/turn-server/udp-connect/main.go
@@ -1,0 +1,105 @@
+// Package main implements a performance optimized TURN server that uses client-specific connected
+// sockets using SO_REUSEPORT.
+package main
+
+import (
+	"context"
+	"flag"
+	"log"
+	"net"
+	"os"
+	"os/signal"
+	"regexp"
+	"strconv"
+	"syscall"
+
+	"github.com/pion/logging"
+	"github.com/pion/turn/v2"
+)
+
+func main() {
+	publicIP := flag.String("public-ip", "", "IP Address that TURN can be contacted by.")
+	port := flag.Int("port", 3478, "Listening port.")
+	users := flag.String("users", "", "List of username and password (e.g. \"user=pass,user=pass\")")
+	realm := flag.String("realm", "pion.ly", "Realm (defaults to \"pion.ly\")")
+	flag.Parse()
+
+	if len(*publicIP) == 0 {
+		log.Fatalf("'public-ip' is required")
+	} else if len(*users) == 0 {
+		log.Fatalf("'users' is required")
+	}
+
+	// Create a UDP listener to pass into pion/turn
+	// pion/turn itself doesn't allocate any UDP sockets, but lets the user pass them in
+	// this allows us to add logging, storage or modify inbound/outbound traffic
+	listenerConfig := &net.ListenConfig{Control: reuseAddr}
+	udpListener, err := listenerConfig.ListenPacket(context.Background(), "udp4",
+		"0.0.0.0:"+strconv.Itoa(*port))
+	if err != nil {
+		log.Panicf("Failed to create TURN server listener: %s", err)
+	}
+
+	// Cache -users flag for easy lookup later
+	// If passwords are stored they should be saved to your DB hashed using turn.GenerateAuthKey
+	usersMap := map[string][]byte{}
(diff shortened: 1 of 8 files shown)


jech:
What's the CPU usage like when running on a single core?

rg0now:
Depends on the traffic rate and the CPU, but in general UDP listeners can handle about 30-60 thousands packets/sec (roughly 100-200 Mbps depending on the packet size) per CPU-core. For comparison, a pure-C TURN can server handle about 100-300 thousand pps and our optimized DPDK-based UDP proxy (no TURN) reaches 10-20 million(!) pps per CPU-core.

jech:
I meant: how much does this patch reduce single-core performance?

rg0now:
It doesn't. Theoretically, it shouldn't, and in practice my super-rudimentary iperf benchmarks confirm this (62500 pps with 20-30 usec mean delay on  localhost  both with and without the patch).

jech:
Ah, cool.  I'll do my best to find time to review your patch, then.

rg0now:
You're going to hate me for this but I think I will double your review load: it seems that we found  a much less intrusive solution, see #295. The performance is a bit smaller but there are no changes in the server code and the required changes seem more secure.

stv0g:
@rg0now Are you okay with it if we close this PR?

rg0now:
> @rg0now Are you okay with it if we close this PR?

Already wanted to suggest this... Please go ahead

Issue #286, opened 2023-01-18
Simplify NewServer()


stv0g:
The cognitive complexity of `NewServer()` is too high (36 > 30).

@rg0now has proposed the following change in #267 which I support:

> I tend to agree with this: `NewServer` was already rather complex in the first place. We've added
> just 6 new lines of new fairly trivial code to handle the permissionHandlers.
> 
> One way to go about this is to separate out the two loop bodies into a separate function each. Namely, the code that sets up PacketConnConfigs and calls the readLoop from https://github.com/pion/turn/blob/371fc35851d91a885a0621b295f8e9a3ed6af2d2/server.go#L68 to https://github.com/pion/turn/blob/371fc35851d91a885a0621b295f8e9a3ed6af2d2/server.go#L85 could go into a new function, say, `runPacketConn`, and the code that sets up the ListenerConfigs from https://github.com/pion/turn/blob/371fc35851d91a885a0621b295f8e9a3ed6af2d2/server.go#L90 to https://github.com/pion/turn/blob/371fc35851d91a885a0621b295f8e9a3ed6af2d2/server.go#L114 could be refactored into a function called `runListener` or similar. Would you be willing to take such a patch?

pristupaanastasia:
Hi, can I take this task?

stv0g:
> Hi, can I take this task?

Sure, a PR would be very welcome :) Maybe just align with @rg0now and his changes in #288.

rg0now:
Hi @pristupaanastasia, thanks for https://github.com/l7mp/turn/pull/1. I think your change is OK. However, I'm not sure it's optimal to base your PR on #288: there is a lot to do on #288 until it gets in a mergeable form and it would be shame to delay your change until that happens. My suggestion would be to rebase your PR on top of pion/turn master and submit it here as an independent patch, and once it's merged I'll rebase @288 on top of your change. Wdyt?

stv0g:
Done 🥳

PR #291, opened 2023-01-22
Rewrite NewServer to reduce cyclomatic complexity

pristupaanastasia:
#### Reference issue

https://github.com/pion/turn/issues/286

Fixes #...

### DIFF
--- a/server.go
+++ b/server.go
@@ -28,8 +28,7 @@ type Server struct {
 	packetConnConfigs  []PacketConnConfig
 	listenerConfigs    []ListenerConfig
 	allocationManagers []*allocation.Manager
-
-	inboundMTU int
+	inboundMTU         int
 }
 
 // NewServer creates the Pion TURN server
@@ -57,7 +56,6 @@ func NewServer(config ServerConfig) (*Server, error) {
 		channelBindTimeout: config.ChannelBindTimeout,
 		packetConnConfigs:  config.PacketConnConfigs,
 		listenerConfigs:    config.ListenerConfigs,
-		allocationManagers: make([]*allocation.Manager, len(config.PacketConnConfigs)+len(config.ListenerConfigs)),
 		nonces:             &sync.Map{},
 		inboundMTU:         mtu,
 	}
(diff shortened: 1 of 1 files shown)


server.go
```diff
@@ -172,6 +150,25 @@ func (s *Server) Close() error {
...
+	allocationManager, err := allocation.NewManager(allocation.ManagerConfig{
+		AllocatePacketConn: addrGenerator.AllocatePacketConn,
+		AllocateConn:       addrGenerator.AllocateConn,
+		PermissionHandler:  permissionHandler,
+		LeveledLogger:      s.log,
+	})
+	if err != nil {
+		return allocationManager, err
+	}
+	s.allocationManagers[i] = allocationManager
```
stv0g, line 168: I am just wondering what happens if the user provides both `listenerConfigs` and `packetConnConfigs`?
Arn't those indices then overwritten by each other? Or are those two settings mutually exclusive?

stv0g: Oh, nevermind. I figured it out..

server.go
```diff
@@ -68,22 +68,11 @@ func NewServer(config ServerConfig) (*Server, error) {
 
 	for i := range s.packetConnConfigs {
 		go func(i int, p PacketConnConfig) {
```
stv0g, line 70: I think its a little nicer to create two new functions for the whole goroutine here.
E.g.  `readPacketConn` and `readListener`.

Those functions should also get rid of the `i` variable. Thats just not a nice way to handle the `allocationManagers`. I would just to a `s.allocationManagers = append(s.allocationManagers, ...)`.

stv0g:
@pristupaanastasia I made some more changes to your PR and squashed it into a single commit.
It should be ready for merging now.

@rg0now Do you mind to have a look at the changes before we merge them?

rg0now:
One thing that strikes me as odd is that we seem to call `createAllocationManager` twice for `ListenerConfigs`, once in NewServer (I guess that's the place where we'll eventually want to call it) [here](https://github.com/pristupaanastasia/turn/blob/012bea32ba4bc424713ba632fb4c10094adb71c8/server.go#L77), and then once again in `readListener` [here](https://github.com/pristupaanastasia/turn/blob/012bea32ba4bc424713ba632fb4c10094adb71c8/server.go#L136). This should append two allocation managers per `ListenerConfig`, which should not cause a problem I guess but it's still curious. 

Apart,  there must be good reason for renaming `errors` to `errs` throughout, I'd be interested to know why (maybe I'm doing something wrong in my code?).

Finally, I guess we can simplify [this](https://github.com/pristupaanastasia/turn/blob/012bea32ba4bc424713ba632fb4c10094adb71c8/server.go#L159) like below, is there a reason why we introduce a separate variable? (I really miss the ternary operator from Go but I guess it's only just me). 

```
if handler == nil {
	handler = DefaultPermissionHandler
}

am, err := allocation.NewManager(allocation.ManagerConfig{
	AllocatePacketConn: addrGenerator.AllocatePacketConn,
	AllocateConn:       addrGenerator.AllocateConn,
	PermissionHandler:  handler,
	LeveledLogger:      s.log,
})
```

Note that I may be wrong on each of these points, haven't done any tests myself.

stv0g:
> One thing that strikes me as odd is that we seem to call createAllocationManager twice for ListenerConfigs, once in NewServer (I guess that's the place where we'll eventually want to call it) [here](https://github.com/pristupaanastasia/turn/blob/012bea32ba4bc424713ba632fb4c10094adb71c8/server.go#L77), and then once again in readListener [here](https://github.com/pristupaanastasia/turn/blob/012bea32ba4bc424713ba632fb4c10094adb71c8/server.go#L136).

Ouch :( Sorry this was a mistake by me. It should only be called once per listener in `NewServer`.
@pristupaanastasia initially called it within the Goroutine which required locking. I moved it outside, so we can safe the lock.

> Apart, there must be good reason for renaming errors to errs throughout, I'd be interested to know why (maybe I'm doing something wrong in my code?).

Agreed. I reverted the change.

> Finally, I guess we can simplify [this](https://github.com/pristupaanastasia/turn/blob/012bea32ba4bc424713ba632fb4c10094adb71c8/server.go#L159) like below, is there a reason why we introduce a separate variable? (I really miss the ternary operator from Go but I guess it's only just me).

Nice catch! No its, not just you. I miss it dearly :D

stv0g (approved):

PR #295, opened 2023-01-30
Implementation of per-client UDP readloops, take 2

rg0now:
This is the second attempt to address #287, see also #288

**Problem:** The specific problem we are trying to address is to scale TURN/UDP server listeners beyond a single CPU. Currently we allocate a single global net.PacketConn per TURN/UDP listener that is shared across all clients. Since there is only a single readLoop to drain the global PacketConn, TURN/UDP listeners are practically limited to a single CPU. See more on this [here](https://github.com/l7mp/stunner/issues/60).

**Prior attempts:** The first draft PR #288 addresses this issue by forking off a separate readloop thread per each client, which allows the server to drain each client connection on a separate CPU thread. In order to implement this in a secure way #288 introduces some fairly intrusive changes to the main TURN server implementation: there is an API change where the user must provide a `Connect` callback that the server calls after a new TURN allocation is created to make the per-client connection, and there is some major refactoring in `internal/server` as well to make this possible. Meanwhile, a security problem is still lurking in the background: without the implementation of per-user allocation quotas it is still possible for a single high-bandwidth attacker to starve the rest of the clients by creating multiple TURN allocations and consuming excess CPU.

**Goal:** This PR takes a completely different approach: the aim is to scale UDP/TURN listeners beyond the single-CPU limit but meanwhile (1) minimize the changes required in the existing pion/turn implementation (the PR requires absolutely no change to the server code) and (2) allow the operator to tightly control the number of CPUs allocated to each TURN/UDP listener.

**Implementation:** The idea is that the caller creates a fixed number of UDP listener sockets that share the same local addr:port by using the `SO_REUSEPORT` socket option and then initializes a separate TURN listener per each socket. Note that when multiple sockets share the same local address with `SO_REUSEPORT` then the kernel will load-balance packets across the sockets by the IP five-tuple hash (see  [here](https://elixir.bootlin.com/linux/latest/source/net/ipv4/udp.c#L415)), which makes sure that each packet of a TURN allocation will be processed by the same readLoop and so the allocation manager doesn't need to be shared across CPU threads. By controlling the number of sockets, the caller can control the number of CPUs allocated to each listener. Since this requires no change at all in the TURN server code (every socket will be handled by pion/turn as a separate UDP/TURN listener), the PR contains only a simple extension of the example server (see `examples/turn-server/mult-threaded-udp/main.go`) to implement the caller-side changes.

The pros are simplicity and control over CPU consumption. The cons are (1) that `SO_REUSEPORT` is not portable (we don't see this as a problem: recall that nothing in the server knows anything about whether `SO_REUSEPORT` is in use, all required changes are on the caller side) and (2) performance is slightly smaller than with #288. This is because hash-based load-balancing used by the kernel to distribute TURN allocations across the CPU threads is prone to an unequal distribution of load across worker threads: this can be taken care of by implementing an adaptive load-balancing algorithm [with eBPF](https://www.kernel.org/doc/html/latest/bpf/prog_sk_lookup.html) (say, using the [power-of-two choices algorithm](https://ieeexplore.ieee.org/document/963420)).

**Performance:** Attached is a script to benchmark the naive TURN server with (`mult-threaded-udp`) and without (`simple`) the patch. Requires the turncat TURN client (you can obtain from  [here](https://github.com/l7mp/stunner/tree/main/cmd/turncat)) in the source dir, plus iperfv2 in the PATH. On a 24-core Intel server and 8 iperf threads, the built-in simple server (`./test2.sh simple 16000000`) produces roughly 41 kpps (thousand packets per second) packet rate and uses about 141% CPU, while the server using 4 `SO_REUSEPORT` sockets and so 4 readLoop threads (`./test.sh mult-threaded-udp 16000000`) produces 121 kpps with 390% CPU usage. Observe that both the performance and the CPU usage is smaller than with #288, this is because of the uneven load-distribution (the difference between the least-loaded thread and the most-loaded thread can be more than 50% in our benchmarks).

Feedback appreciated.
[test2.sh.gz](https://github.com/pion/turn/files/10535074/test2.sh.gz)

### DIFF
--- a/examples/turn-server/simple-multithreaded/main.go
+++ b/examples/turn-server/simple-multithreaded/main.go
@@ -1,0 +1,111 @@
+//go:build !wasm
+
+// Package main implements a multi-threaded TURN server
+package main
+
+import (
+	"context"
+	"flag"
+	"log"
+	"net"
+	"os"
+	"os/signal"
+	"regexp"
+	"strconv"
+	"syscall"
+
+	"github.com/pion/turn/v2"
+	"golang.org/x/sys/unix"
+)
+
+func main() {
+	publicIP := flag.String("public-ip", "", "IP Address that TURN can be contacted by.")
+	port := flag.Int("port", 3478, "Listening port.")
+	users := flag.String("users", "", "List of username and password (e.g. \"user=pass,user=pass\")")
+	realm := flag.String("realm", "pion.ly", "Realm (defaults to \"pion.ly\")")
+	threadNum := flag.Int("thread-num", 1, "Number of server threads (defaults to 1)")
+	flag.Parse()
+
+	if len(*publicIP) == 0 {
+		log.Fatalf("'public-ip' is required")
+	} else if len(*users) == 0 {
+		log.Fatalf("'users' is required")
+	}
+
+	addr, err := net.ResolveUDPAddr("udp", "0.0.0.0:"+strconv.Itoa(*port))
+	if err != nil {
+		log.Fatalf("Failed to parse server address: %s", err)
+	}
+
+	// Cache -users flag for easy lookup later
+	// If passwords are stored they should be saved to your DB hashed using turn.GenerateAuthKey
+	usersMap := map[string][]byte{}
+	for _, kv := range regexp.MustCompile(`(\w+)=(\w+)`).FindAllStringSubmatch(*users, -1) {
+		usersMap[kv[1]] = turn.GenerateAuthKey(kv[1], *realm, kv[2])
+	}
+
+	// Create `numThreads` UDP listeners to pass into pion/turn
+	// pion/turn itself doesn't allocate any UDP sockets, but lets the user pass them in
+	// this allows us to add logging, storage or modify inbound/outbound traffic
+	// UDP listeners share the same local address:port with setting SO_REUSEPORT and the kernel
+	// will load-balance received packets per the IP 5-tuple
+	listenerConfig := &net.ListenConfig{
+		Control: func(network, address string, conn syscall.RawConn) error {
+			var operr error
+			if err = conn.Control(func(fd uintptr) {
+				operr = syscall.SetsockoptInt(int(fd), syscall.SOL_SOCKET, unix.SO_REUSEPORT, 1)
+			}); err != nil {
+				return err
+			}
+
+			return operr
+		},
+	}
+
+	relayAddressGenerator := &turn.RelayAddressGeneratorStatic{
+		RelayAddress: net.ParseIP(*publicIP), // Claim that we are listening on IP passed by user
+		Address:      "0.0.0.0",              // But actually be listening on every interface
+	}
+
+	packetConnConfigs := make([]turn.PacketConnConfig, *threadNum)
+	for i := 0; i < *threadNum; i++ {
+		conn, listErr := listenerConfig.ListenPacket(context.Background(), addr.Network(), addr.String())
+		if listErr != nil {
+			log.Fatalf("Failed to allocate UDP listener at %s:%s", addr.Network(), addr.String())
+		}
+
+		packetConnConfigs[i] = turn.PacketConnConfig{
+			PacketConn:            conn,
+			RelayAddressGenerator: relayAddressGenerator,
+		}
+
+		log.Printf("Server %d listening on %s\n", i, conn.LocalAddr().String())
+	}
+
+	s, err := turn.NewServer(turn.ServerConfig{
+		Realm: *realm,
+		// Set AuthHandler callback
+		// This is called every time a user tries to authenticate with the TURN server
+		// Return the key for that user, or false when no user is found
+		AuthHandler: func(username string, realm string, srcAddr net.Addr) ([]byte, bool) {
+			if key, ok := usersMap[username]; ok {
+				return key, true
+			}
+			return nil, false
+		},
+		// PacketConnConfigs is a list of UDP Listeners and the configuration around them
+		PacketConnConfigs: packetConnConfigs,
+	})
+	if err != nil {
+		log.Panicf("Failed to create TURN server: %s", err)
+	}
+
+	// Block until user sends SIGINT or SIGTERM
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+	<-sigs
+
+	if err = s.Close(); err != nil {
+		log.Panicf("Failed to close TURN server: %s", err)
+	}
+}
--- a/examples/README.md
+++ b/examples/README.md
@@ -31,6 +31,9 @@ You could also intercept these reads/writes if you want to filter traffic going 
 #### simple
 This example is the most minimal invocation of a Pion TURN instance possible. It has no custom behavior, and could be a good starting place for running your own TURN server.
 
+#### simple-multithreaded
+A multithreaded version of the `simple` Pion TURN server, demonstrating how to scale a Pion UDP TURN server to multiple CPU cores. By default, Pion TURN servers use a single UDP socket that is shared across all clients, which limits Pion UDP/TURN servers to a single CPU thread. This example passes a configurable number of UDP sockets to the TURN server, which share the same local `address:port` pair using the `SO_REUSEPORT` socket option. This then lets the server to create a separate readloop to drain each socket. The OS kernel will distribute packets received on the `address:port` pair across the sockets by the IP 5-tuple, which makes sure that all packets of a TURN allocation will be correctly processed in a single readloop.
+
 #### tcp
 This example demonstrates listening on TCP. You could combine this example with `simple` and you will have a Pion TURN instance that is available via TCP and UDP.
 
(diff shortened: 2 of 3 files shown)


stv0g:
Sorry @rg0now, that I didnt had time to review the PR yet.
I will do so later today or tomorrow. But I am currious, which of the two approaches do you prefer?

rg0now:
> But I am curious, which of the two approaches do you prefer?

Frankly, I would opt for this one, just because it is much less intrusive and more secure. The other PR requires some massive reorg all  around the place plus it is still susceptible to eating up all CPUs. At some point  we may consider implementing per-user quotas (see the [note](https://github.com/l7mp/turn/blob/e3ad12cdd9125f35867a36f622aa6f512b97621c/internal/server/turn.go#L109)), which would fix the security issue at least, but I'm still uncomfortable with the massive change.

I've added a note to examples/README.md to point users to the new example and also applied some minor stylistic changes throughout.

stv0g:
Currently, the WASM pipeline fails as the new multihreaded examples as excluded all files from the package via build-tags.
I suggest you just add an empty file into the directory which is not excluded.

https://github.com/pion/turn/actions/runs/4078041150/jobs/7027770548

examples/turn-server/simple-multithreaded/main.go
```diff
@@ -0,0 +1,102 @@
...
+	packetConnConfigs := make([]turn.PacketConnConfig, *threadNum)
+	for i := 0; i < *threadNum; i++ {
+		conn, listErr := listenerConfig.ListenPacket(context.Background(), addr.Network(), addr.String())
+		if listErr != nil {
+			log.Fatalf("Failed to allocate UDP listener at %s:%s", addr.Network(), addr.String())
+		}
+
+		packetConnConfigs[i] = turn.PacketConnConfig{
+			PacketConn: conn,
+			RelayAddressGenerator: &turn.RelayAddressGeneratorStatic{
```
stv0g, line 67: I think we should be able to pass the same `turn. RelayAddressGeneratorStatic` to all the listener configs?
I would say, just declare it outside the loop and pass the same pointer to each listener config.

examples/turn-server/simple-multithreaded/main.go
```diff
@@ -0,0 +1,102 @@
...
+
+	if len(*publicIP) == 0 {
+		log.Fatalf("'public-ip' is required")
+	} else if len(*users) == 0 {
+		log.Fatalf("'users' is required")
+	}
+
+	addr, err := net.ResolveUDPAddr("udp", "0.0.0.0:"+strconv.Itoa(*port))
+	if err != nil {
+		log.Fatalf("Failed to parse server address: %s", err.Error())
```
stv0g, line 35: Please remove `.Error()` as its not required.

examples/turn-server/simple-multithreaded/main.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// Create `numThreads` UDP listeners to pass into pion/turn
+	// pion/turn itself doesn't allocate any UDP sockets, but lets the user pass them in
+	// this allows us to add logging, storage or modify inbound/outbound traffic
+	// UDP listeners share the same local address:port with setting SO_REUSEPORT and the kernel
+	// will load-balance received packets per the IP 5-tuple
+	listenerConfig := &net.ListenConfig{Control: func(network, address string, conn syscall.RawConn) error {
+		return conn.Control(func(descriptor uintptr) {
+			if optErr := syscall.SetsockoptInt(int(descriptor), syscall.SOL_SOCKET, unix.SO_REUSEPORT, 1); optErr != nil {
+				log.Fatalf("Failed to set SO_REUSEPORT")
```
stv0g, line 53: Please pass through the `optErr`

examples/turn-server/simple-multithreaded/main.go
```diff
@@ -0,0 +1,102 @@
...
+			}
+		})
+	}}
+
+	packetConnConfigs := make([]turn.PacketConnConfig, *threadNum)
+	for i := 0; i < *threadNum; i++ {
+		conn, listErr := listenerConfig.ListenPacket(context.Background(), addr.Network(), addr.String())
+		if listErr != nil {
+			log.Fatalf("Failed to allocate UDP listener at %s:%s", addr.Network(), addr.String())
```
stv0g, line 62: Please pass through the `listErr`

examples/turn-server/simple-multithreaded/main.go
```diff
@@ -0,0 +1,102 @@
...
+			if key, ok := usersMap[username]; ok {
+				return key, true
+			}
+			return nil, false
+		},
+		// PacketConnConfigs is a list of UDP Listeners and the configuration around them
+		PacketConnConfigs: packetConnConfigs,
+	})
+	if err != nil {
+		log.Panic(err)
```
stv0g, line 91: Can we add a message like:

```go
log.Panicf("Failed to create TURN server: %s", err)
```

examples/turn-server/simple-multithreaded/main.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	// Block until user sends SIGINT or SIGTERM
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+	<-sigs
+
+	if err = s.Close(); err != nil {
+		log.Panic(err)
```
stv0g, line 100: Same here as with the previous comment :)

stv0g (changes requested):
I agree with your assessment regarding the two `readLoops` PRs :)

Also great work on this one to keep it simple and concise 👍 

I am just wondering, if we maybe want to include this not just as an example but into the package itself. Do you have an opinion on that?

I fear that this might kick-off a whole bunch of server variants. Each with its own `turn.NewXXXServer()` function and `turn.XXXServerConfig` type.

Thats something, I would like to avoid to keep it maintainable..

examples/turn-server/simple-multithreaded/main.go
rg0now: I did this but, frankly, I found the end result (taken from https://iximiuz.com/en/posts/go-net-http-setsockopt-example/) lacking in may ways. I can't believe this API is so strange...

examples/turn-server/simple-multithreaded/main.go
rg0now: I'm not sure this is doable, we are already in the `main()` scope so we cannot just return an error. I guess failing to create a listener sockets warrants a fatal error at this point so I left this as is.

rg0now:
> Currently, the WASM pipeline fails as the new multihreaded examples as excluded all files from the package via build-tags.
> I suggest you just add an empty file into the directory which is not excluded.

I wish I understood this but I can't. I guess the CI errs due to the new soft-dependency `golang.org/x/sys`, is this somehow incompatible with WASM (would make sense)? where should I put an empty file?

I've started to think about whether the whole hassle is worth the price: I mean, adding a new dependency (which is not even portable) just so that we can demonstrate a use case that may not be too interesting outside a small audience somehow seems wrong. Maybe an example is an overkill? Should I just post a gist and hope Google picks it up? Wdyt?

> I am just wondering, if we maybe want to include this not just as an example but into the package itself. Do you have an opinion on that?
> I fear that this might kick-off a whole bunch of server variants. Each with its own turn.NewXXXServer() function and turn.XXXServerConfig type.

I'm not sure we want to touch the API and/or `turn.Server` per se: it seems that we get reasonable performance without touching the code in any way, just by calling the existing API in a weird-ish way. Plus this is applicable only to UDP (TCP, TLS and DTLS spawn per-allocation readLoops), so I guess there's no need to add the specific `turn.NewXXXServer()` functions. Wdyt?

stv0g:
> I'm not sure we want to touch the API and/or turn.Server per se

I agree. Lets merge it like this.

> I wish I understood this but I can't. I guess the CI errs due to the new soft-dependency golang.org/x/sys, is this somehow incompatible with WASM (would make sense)?

Please try to add a `//go:build !wasm` to the `main.go` of your new example. I dont think we should attempt to build this for WASM

> I've started to think about whether the whole hassle is worth the price: I mean, adding a new dependency (which is not even portable) just so that we can demonstrate a use case that may not be too interesting outside a small audience somehow seems wrong. Maybe an example is an overkill? Should I just post a gist and hope Google picks it up? Wdyt?

I think its worth the effort. 👍

rg0now:
> Please try to add a //go:build !wasm to the main.go of your new example. I dont think we should attempt to build this for WASM

Done. Thanks a lot for the review work, it's extremely useful, I learn a lot.

stv0g:
Seems like your last change is missing some entries in `go.sum`:

```
stv0g@ubuntu-linux-22-04-desktop:~/workspace/pion/turn$ go test ./...
go: golang.org/x/sys@v0.4.0: missing go.sum entry; to add it:
	go mod download golang.org/x/sys
go: downloading github.com/pion/transport/v2 v2.0.2
go: golang.org/x/sys@v0.4.0: missing go.sum entry; to add it:
```

Can you run `go mod tidy` and push again? Thanks a lot.

rg0now:
I'm terribly sorry, hope this version will be fine now.

stv0g (approved):

stv0g:
Thanks a lot :)

Issue #293, opened 2023-01-23
TestServer/Filter_on_client_address_and_peer_IP is failing


enobufs:
### Your environment.
  - Version: 310bd86 (current master/HEAD
  - Browser: n/a
  - Other Information - macOS Monterey M1, macOS BigSur x86

### What did you do?
go test -v -p 1 ./...

### What did you expect?
It should pass all tests.

### What happened?
```
=== RUN   TestServer/Filter_on_client_address_and_peer_IP
    server_test.go:148:
        	Error Trace:	/Users/ytakeda/Projects/pion/ws/turn/server_test.go:148
        	Error:      	Received unexpected error:
        	            	listen udp4 127.0.0.33:54321: bind: can't assign requested address
        	Test:       	TestServer/Filter_on_client_address_and_peer_IP
    server_test.go:159:
        	Error Trace:	/Users/ytakeda/Projects/pion/ws/turn/server_test.go:159
        	Error:      	Received unexpected error:
        	            	turn: conn cannot not be nil
        	Test:       	TestServer/Filter_on_client_address_and_peer_IP
--- FAIL: TestServer (0.21s)
    --- PASS: TestServer/simple (0.00s)
    --- PASS: TestServer/default_inboundMTU (0.00s)
    --- PASS: TestServer/Set_inboundMTU (0.00s)
    --- FAIL: TestServer/Filter_on_client_address_and_peer_IP (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x0 pc=0x1024a3298]

goroutine 62 [running]:
testing.tRunner.func1.2({0x102541040, 0x1026b3f50})
	/usr/local/go/src/testing/testing.go:1396 +0x1c8
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1399 +0x378
panic({0x102541040, 0x1026b3f50})
	/usr/local/go/src/runtime/panic.go:884 +0x204
github.com/pion/turn/v2.(*Client).Listen(0x102579098?)
	/Users/ytakeda/Projects/pion/ws/turn/client.go:165 +0x18
github.com/pion/turn/v2.TestServer.func4(0x0?)
	/Users/ytakeda/Projects/pion/ws/turn/server_test.go:160 +0x350
testing.tRunner(0x14000083040, 0x140000aa078)
[...]

PR #297, opened 2023-01-31
fix PermissionHandler test failure on MacOS, see #293

rg0now:
The test client binds to 127.0.0.33 but this localhost alias is not up by default on MacOS. This fix reverts the test client to 127.0.0.1 that should be up by default.

### DIFF
--- a/server_test.go
+++ b/server_test.go
@@ -133,7 +133,7 @@ func TestServer(t *testing.T) {
 						Address:      "0.0.0.0",
 					},
 					PermissionHandler: func(src net.Addr, peer net.IP) bool {
-						return src.String() == "127.0.0.33:54321" &&
+						return src.String() == "127.0.0.1:54321" &&
 							peer.Equal(net.ParseIP("127.0.0.4"))
 					},
 				},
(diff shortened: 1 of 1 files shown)


stv0g (approved):

stv0g:
Thanks for fixing it so quickly @rg0now :)

Can you fix the commit message and merge it?

rg0now:
> Thanks for fixing it so quickly @rg0now :)

I hope this really fixes it, I'm on Linux, cannot test on MacOS. Have you heard back from @enobufs, did he/she manage to test the fix? 

> Can you fix the commit message and merge it?

As per the commit message, would this work: `Fix PermissionHandler test failure on MacOS, closes #293`? As per merging it: I'm afraid I don't have the rights to do that...;-)

stv0g:
> As per the commit message, would this work: Fix PermissionHandler test failure on MacOS, closes #293?

Unfortunately not. Here are the rules for the commit message:
```
Error: Separate subject from body with a blank line

The preceding commit message is invalid
it failed 'Separate subject from body with a blank line' of the following checks

* Capitalize the subject line
```

See: https://github.com/pion/turn/actions/runs/4055750593/jobs/6979306734

rg0now:
Sorry, I should have learned this by now. I've force-pushed a new commit (see https://github.com/l7mp/turn/commit/61ba72a5993b0049d6841463ef86edae15ae7551) but it doesn't seem to show up here for some reason. Maybe GitHub is slow to sync stuff today? Or maybe I cannot change approved PRs?

stv0g:
Thanks a lot @rg0now  :)

PR #298, opened 2023-02-02
Add benchmark test case for TURN/UDP profiling

rg0now:
A new benchmark test is added that allows to stress the TURN/UDP client and the server in a simple setup. This allows to get an insight into the running times of the main TURN/UDP client-server routines with real I/O.

Usage:
- run: `go test -cpuprofile cpu.prof -bench=. -run=XXX -benchtime=10s`
- view profile: `go tool pprof -http=":8001" ./cpu.prof &`

### DIFF
--- a/server_test.go
+++ b/server_test.go
@@ -4,6 +4,7 @@
 package turn
 
 import (
+	"fmt"
 	"net"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


rg0now:
Overall, it seems pion/turn is doing fairly well in the benchmarks: judging from the flame graph (click View/Flame-graph in the GUI), roughly 65% of the running time is spent in I/O (`UDPConn.WriteTo/ReadFrom`), 20% is taken by the Go runtime and only the remaining 15% is spent on TURN client/server packet forging and business logics. I was particularly happy to not see excess packet-buffer/byte-slice copies all around the place, which is a frequent source of overhead
udpaddr_string
 in networking code.

One minor outlier is `UDPAddr.String()`: we spend almost 7% of the total running time stringifying UDP addresses. The major user is the allocation manager's `GetAllocation()` routine (>4% total running time) that calls `Fivetuple.FingerPrint()` per each received packet to find the allocation in the allocation map, effectively calling `UDPAddr.Stringify()` per each packet multiple times (see attached screenshot). I'm wondering whether we could (and should) try to optimize this.

Apart from this, however, pion/turn seems to be doing nicely and there do not seem to be major low-hanging optimization targets.

stv0g:
> One minor outlier is UDPAddr.String()

I think we should modernize the code and use the `netip` package here. We could use its `AddrPort` type to make the `FiveTuple` comparable and hence get rid of the stringification completely :)

stv0g:
And, I fixed some of the linter warnings which you silcenced via `//nolint`

rg0now:
Thanks! I wish you didn't have to do all the extra work with my PRs, I'll try to stick the coding style.

As per `netip.AddrPort`: I guess the definition of `FiveTuple` would change as follows:

```
type FiveTuple struct {
	Protocol
	SrcAddr, DstAddr netip.AddrPort
}
```

Now I see that `netip.AddrPort` is safe for comparison, but is the struct it is embedded in also comparable? I mean, with the new definition can we write this?

```
// Equal asserts if two FiveTuples are equal
func (f *FiveTuple) Equal(b *FiveTuple) bool {
	return *f == *b
}
```

And then, can we also use the `FiveTuple` directly as a map key?

```
// Manager is used to hold active allocations
type Manager struct {
...
	allocations  map[FiveTuple]*Allocation
...
}

// GetAllocation fetches the allocation matching the passed FiveTuple
	return m.allocations[*fiveTuple]
}
```

My understanding is that this all would be OK, but I've never used `netip` nor struct-as-map-key myself.

And if this is all OK, are you willing to take a PR that makes this change? Is it OK to add a new package dependency?

stv0g:
> Thanks! I wish you didn't have to do all the extra work with my PRs, I'll try to stick the coding style.

Please do not worry about that. I am really happy to have the chance to do this review work.
I started to like it, as it allows me to learn about the Pion code-base step-by-step :)

PR #300, opened 2023-02-04
Port old fuzzing code to Go's new fuzzing tools

### LINKED ISSUE #292: Port old fuzzing code to Go's new fuzzing tools
As introduced by Go 1.18

See: https://go.dev/security/fuzz/

stv0g:
Closes #292


ashellunts (approved):
There are new functions added, but I don't see them used. Will they be used in dependent projects? Also would you mind adding tests for the new functions?

stv0g:
Thanks for the review :)

> There are new functions added, but I don't see them used. Will they be used in dependent projects? Also would you mind adding tests for the new functions?

I actually added them for the fuzzing test to have full coverage of all TURN attributes.

PR #302, opened 2023-02-13
Make type assertion option in util.go

vernonr3:
The test functions use the net library internal functions to create examples of UDPAddr and TCPAddr for interpretation.

Provide unit tests for the new functions to validate the interpretation of UDP and TCP addresses and ports

Provide an example of the mocking that now becomes possible. Whilst the "typecast" appears neat it prevents test code from providing it's own mocked contents from the Network() and String() functions that net.Addr requires.

This means testing code may not be able to easily spoof IP addresses or ports of interest. It would appear to severely limit the ability to test TURN code in isolation as the server internals rely on this function - and link them to the request packet and destination address.

The single global variable is there to provide a big switch - between the more testable variant and the existing faster code.

#### Reference issue
Fixes #...

### DIFF
--- a/internal/ipnet/util.go
+++ b/internal/ipnet/util.go
@@ -2,14 +2,56 @@
 package ipnet
 
 import (
+	"encoding/hex"
 	"errors"
 	"net"
+	"strconv"
 )
 
 var errFailedToCastAddr = errors.New("failed to cast net.Addr to *net.UDPAddr or *net.TCPAddr")
 
-// AddrIPPort extracts the IP and Port from a net.Addr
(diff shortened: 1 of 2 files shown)


stv0g:
@vernonr3 Please try to avoid all caps titles :) Title-case (see above) is easier to read

vernonr3:
> @vernonr3 Please try to avoid all caps titles :) Title-case (see above) is easier to read

I agree - sorry - I was trying to follow what I believed what requested in lint-commit-message.sh. -)

stv0g:
I currently see big value in this PR and would propose to close it.

PR #304, opened 2023-02-22
Use netip.AddrPort for FiveTuple, closes #299

### LINKED ISSUE #299: Use netip.AddrPort for FiveTuple to reduce lookup times
Thanks to @rg0now new benchmarks (see #298) we have some number about the cost of the current allocation lookup:

> One minor outlier is UDPAddr.String(): we spend almost 7% of the total running time stringifying UDP addresses. The major user is the allocation manager's GetAllocation() routine (>4% total running time) that calls Fivetuple.FingerPrint() per each received packet to find the allocation in the allocation map, effectively calling UDPAddr.Stringify() per each packet multiple times (see attached screenshot). I'm wondering whether we could (and should) try to optimize this.

I think we should modernize the code and use the netip package here. We could use its AddrPort type to make the FiveTuple comparable and hence get rid of the stringification completely :)

> As per `netip.AddrPort`: I guess the definition of FiveTuple would change as follows:
> 
> ```go
> type FiveTuple struct {
> 	Protocol
> 	SrcAddr, DstAddr netip.AddrPort
> }
> ```

Yes exactly.

> Now I see that `netip.AddrPort` is safe for comparison, but is the struct it is embedded in also comparable?

Yes. The [language spec says](https://go.dev/ref/spec#Comparison_operators):

> Struct types are comparable if all their field types are comparable. Two struct values are equal if their corresponding non-[blank](https://go.dev/ref/spec#Blank_identifier) field values are equal. The fields are compared in source order, and comparison stops as soon as two field values differ (or all fields have been compared).

> I mean, with the new definition can we write this?
> 
> ```go
> // Equal asserts if two FiveTuples are equal
> func (f *FiveTuple) Equal(b *FiveTuple) bool {
> 	return *f == *b
> }
> ```

Yes we can. But I think an `Equal()` function is not even required here.

> And then, can we also use the FiveTuple directly as a map key?

Yes, as all the fields of `FiveTuple` are comparable themself.

```go
// Manager is used to hold active allocations
type Manager struct {
...
	allocations  map[FiveTuple]*Allocation
...
}
```

```go
// GetAllocation fetches the allocation matching the passed FiveTuple
	return m.allocations[*fiveTuple]
}
```

> My understanding is that this all would be OK, but I've never used netip nor struct-as-map-key myself.

Yes :)

> And if this is all OK, are you willing to take a PR that makes this change? Is it OK to add a new package dependency?

Yes. I created this new issue to track the PR.

rg0now:
This commit refactors FiveTuple to use `netip.AddrPort` instead of `net.Addr` to store the source and the destination address in order to improve performance on frequent GetAllocations.

#### Description

The `GetAllocation()` routine calls `Fivetuple.FingerPrint()` per each received packet to find the allocation in the allocation map, effectively calling `UDPAddr.String()` per each packet multiple times, adding 4-5% CPU overhead per packet.  This commit refactors FiveTuple to use `netip.AddrPort` instead of `net.Addr` to store the source and the destination address: since `netip.AddrPort` is comparable we can use it directly as a map key instead of having to stringify the addresses first. As the FiveTuple.SrcAddr (being now a `netip.AddrPort`) can no longer be passed directly to `net.WriteTo()` (as `netip.AddrPort` does not implement `Network()` but even if it did `transport.VNet` and company still require a `net.UDPAddr`), the client address is now directly stored in the allocation.

This commit should cause no user-visible change.

#### Benchmark

Benchmarks (use `go test -cpuprofile cpu.prof -bench=. -run=XXX -benchtime=10s`) are somewhat inconclusive; see the attached CPU profiles against the current master and the refactored TURN server:

```
go tool pprof -http=":8001" ./cpu-master &
go tool pprof -http=":8002" ./cpu-refactor &
```

On the one hand, current master spends about >8sec in `GetAllocation`, this all but disappears (<.2 sec) with the refactored version. We get some new overhead as netip.AddrPort seems to perform a malloc in the hot path but this does not seem like a massive loss. At the same time, the overall benchmark results do not seem to show significant gains, but this may be due to that we benchmark the client (that this PR didn't touch) and the server (which it did) jointly.

The bad news is that `UDP.String()` still appears in the CPU profile: it seems that the spend quite some time to stringify the client and server addresses for `Log.Debug` [here](https://github.com/pion/turn/blob/b0f8b534a5d8018dec8749f494e428223958a047/internal/server/server.go#L36) and [here](https://github.com/pion/turn/blob/b0f8b534a5d8018dec8749f494e428223958a047/internal/server/server.go#L46), just to drop the result  anyway unless we run at the debug loglevel in production (which is not really useful); see also this [pion/logging issue](https://github.com/pion/logging/issues/10).

#### Reference issue
Fixes #299

[cpu-master.prof.gz](https://github.com/pion/turn/files/10805994/cpu-master.prof.gz)
[cpu-refactor.prof.gz](https://github.com/pion/turn/files/10805995/cpu-refactor.prof.gz)


internal/allocation/allocation.go
```diff
@@ -52,8 +52,9 @@ func addr2IPFingerprint(addr net.Addr) string {
 }
 
 // NewAllocation creates a new instance of NewAllocation.
-func NewAllocation(turnSocket net.PacketConn, fiveTuple *FiveTuple, log logging.LeveledLogger) *Allocation {
+func NewAllocation(turnSocket net.PacketConn, clientAddr net.Addr, fiveTuple *FiveTuple, log logging.LeveledLogger) *Allocation {
```
stv0g, line 55: Whats your motivation to pass the `clientAddr` separately instead of taking it from `fiveTuple`?

I think we could add some getters to the `FiveTuple` which provides a `net.{TCP,UDP}Addr` depending on `FiveTuple.Protocol`.

stv0g: Here is an example:

```go
// SrcAddr returns the tuples source address as net.Addr
func (t FiveTuple) SrcAddr() net.Addr {
	switch t.protocol {
	case UDP:
		return &net.UDPAddr{
			IP:   t.srcAddr.Addr().AsSlice(),
			Port: int(t.srcAddr.Port()),
		}
	case TCP:
		return &net.TCPAddr{
			IP:   t.srcAddr.Addr().AsSlice(),
			Port: int(t.srcAddr.Port()),
		}
	default:
		return nil
	}
}
```

stv0g (changes requested):
This looks good to me with the exception of passing the client address separately.

There has been a discussion in the Pion Slack recently about whether we really want to raise the minimal Go version to 1.18 (which introduced the `netip` package).
The general opinion seems to be, that we should raise the minimal version.

However, there are some concerns that this means that we can not build Pion code with the Go compilers shipped in Debian/Ubuntu stable/Bullseye. We might wait until later this year when the next Debian stable is released with Go 1.19..

Connected to this, I realised that there is also an `internal/netip` package which we should probably attempt to eliminate with this PR as well.

internal/allocation/allocation.go
rg0now: This was *exactly* the approach I implemented first but then it occurred to me that this way we would do *two* address type conversions per each packet: first from `net.UDPAddr`/`net.TCPAddr` to `netip.AddrPort` when we build the `FiveTuple` to call `GetAllocation`, and then back again to `net.UDPAddr`/`net.TCPAddr` in the `SrcAddr` function to pass the definition address to`WriteTo`. Now I don't know how costly these conversions are but I thought that if we can get away with just one then this would buy us some extra performance, and storing the original source address as a `net.UDPAddr`/`net.TCPAddr` in the allocation allows us to do that.

That being said, your code is still significantly simpler so if you insist I'll be happy to implement it. Wdyt?

stv0g: Good question. I am not sure about it.. It could be a case of premature optimization. So to have a conclusive decision we should probably benchmark it... But in general I would prefer readability over performance if we can not show the performance benefits.

rg0now: I think if readability is number one priority (I can relate to this POV) then we should drop this PR all together. It already decreases readability by a large margin and I could not show massive performance improvement. I don't say there's no improvement (we remove some 4-7% overhead from `GetAllocation`), all I'm saying is that the end-to-end tests do not seem to show significant speedup. 

Conclusion: we are I/O bounded, so optimizing code will not help...

stv0g:
Hi @rg0now,

Lets close the PR for now. It seems like Pion, will keep compatability with Go < 1.18 for now until major Distros have updated their Go compiler to a newer version. Hence, we can not utilize `netip` (yet).

Lets keep #299 open for now and revisit it later once we switch to Go 1.18 and newer.

PR #305, opened 2023-02-22
Ignore "use of closed network connection" errors

rg0now:
Every time the server returns a response to the client it calls `BuildAndSend()`, which assumes that the other side is still listening and returns a "use of closed network connection" error when it does not. This error always shows up when the client disconnects first, which is totally normal, but messes up logs and fires stale alarms. This commit silences these errors.

I am fairly certain that this change does not break anything. Yet, please give feedback.

### DIFF
--- a/internal/server/util.go
+++ b/internal/server/util.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"crypto/md5" //nolint:gosec,gci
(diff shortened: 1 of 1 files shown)


stv0g (approved):
Looks good to me :)

PR #306, opened 2023-02-23
Close accepted listener connections

stv0g:
This is a new PR for the issue originally addressed in #232


server.go
```diff
@@ -122,30 +120,39 @@ func (s *Server) Close() error {
...
+			return
+		}
+
+		go func() {
+			s.readLoop(NewSTUNConn(conn), am)
 
-	if err := am.Close(); err != nil {
-		s.log.Errorf("Failed to close AllocationManager: %s", err)
+			if err := conn.Close(); err != nil {
+				s.log.Errorf("failed to close conn: %s", err)
```
rg0now, line 135: I'm not sure we want to report all these `"failed to close conn` errors: most of the time the connection is already closed by the time we get here, no? Won't this yield lots of useless errors?

stv0g: We are now only showing an error if its not an `net.ErrClosed`

server.go
```diff
@@ -65,21 +66,19 @@ func NewServer(config ServerConfig) (*Server, error) {
...
 		}
-
-		go s.readPacketConn(cfg, am)
 	}
 
 	for _, cfg := range s.listenerConfigs {
-		am, err := s.createAllocationManager(cfg.RelayAddressGenerator, cfg.PermissionHandler)
-		if err != nil {
-			return nil, fmt.Errorf("failed to create AllocationManager: %w", err)
+		if err := s.startReadLoop(cfg.RelayAddressGenerator, cfg.PermissionHandler, func(am *allocation.Manager) {
```
rg0now, line 77: I perfectly understand why the callback is necessary here but do we really need to complicate our lives that much just to save on calling `createAllocationManager` twice (once in the `PackerConnConfig` loop and then in the `ListenerConfig` loop)? My head hurts every time I try to understand what's going on here. I mean it's elegant and nice and we agreed to make it this way but somehow I feel that this level of generalization hurts readabiliy.

Just my 2 cents of course.

rg0now (approved):

server.go
stv0g: I fully agree. This reduced readability. I've taken the DRY principle too far here.

I reverted it.

stv0g:
@rg0now Would you approve and merge the PR for me? Thanks :)

rg0now (approved):
This version is now clear and readable, thanks!

stv0g:
Thanks @rg0now for the review. But apparently I still can not merge the PR :(

@at-wat do you mind approving it as well?

at-wat (approved):
LGTM

PR #309, opened 2023-02-28
Allow Go tracing to be added without degrading production speed

vernonr3:
The internal/profiling files provide an profiling interface so that code gets included if and only if -tags debug is stated at compile / test time. The server.go opens and closes the trace. It also pushes the profiling interface to areas of concern by piggybacking it inside the server & request structures 
Using the tracing interface in internal/server/server.go shows the use of user regions. This gives a developer more focused analysis of the timing and other issues occurring in practice.

#### Description
The internal profiling files (production_stub.go and profiling.go) are paired. The stub has //go:build !debug and the active profiling.go has //go:build debug. This allows conditional compilation based on a -tags debug

So if testing a benchmark with tracing the command line would be

go test -tags debug -benchmem -run=^$ -bench ^BenchmarkServer$ github.com/pion/turn/v2

If testing without tracing simply omit the -tags debug i.e.
go test -benchmem -run=^$ -bench ^BenchmarkServer$ github.com/pion/turn/v2

The standard output file is trace.out. An alternative name can be set in the OpenTracing call

The go tool trace trace.out will parse and analyse the trace file and launch a local server to display results.

The example files were generated on a older i7 8 core machine using the RunBenchmarkServer in server_test.go. My suspicion is that these aren't showing a true picture of performance because they don't represent a realistic workload for a TURN server. (lots of tiny data packets)

TraceExample
GoRoutineExample
SchedulerExample
UserRegionExample

#### Reference issue
Fixes #...

### DIFF
--- a/internal/profiling/production_stub.go
+++ b/internal/profiling/production_stub.go
@@ -4,22 +4,26 @@ package profiling
 
 import (
 	"fmt"
-
 	"runtime/trace"
 
 	"github.com/pion/logging"
 )
 
+// SayHello just provides easy indication of which version has been included
 func SayHello() {
 	fmt.Println("welcome to production")
 }
 
+// NewProfiling provides a stub function when in production, giving back an empty struct
 func NewProfiling(filename string, mLogger logging.LeveledLogger) *Profiling {
 	return &Profiling{}
 }
 
+// OpenTracing provides a stub function when in production
 func (p *Profiling) OpenTracing(filename string, topLevel string) {}
 
+// CloseTracing provides a stub function when in production
 func (p *Profiling) CloseTracing() {}
 
-func (p *Profiling) SetRegion(regionname string) *trace.Region {}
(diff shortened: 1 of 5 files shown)


vernonr3:
Paused - further testing required...

rg0now:
I really love this. Can you include a profiling output sample for a teaser?

vernonr3:
Changed logic slightly within the SetRegion so the returned value is kept within the struct, and the validity of this return value can be tested before calling the End function.

stv0g:
I think this is to generic to be included in the Turn package. How about we create a separate package for this?

vernonr3:
> I think this is to generic to be included in the Turn package. How about we create a separate package for this?

Its a fair point. It could certainly be useful in other packages within the pion "group". 
How do we best ensure that people using a particular package like Turn are fully aware of it?

stv0g:
Outdated

PR #314, opened 2023-03-21
Fix reused loop counter in the server benchmark

rg0now:
RunBenchmarkServer uses two embedded loops to stress-test the server: an outer loop to make just as many iterations as required for the requested benchmark runtime, and an inner loop that walks through the clients and sends one package from each. Formerly, both loops used the same loop counter variable which led to incorrect results. This change fixes this issue.


stv0g (approved):
Thanks :)

PR #148, opened 2020-07-29
RFC6062 TURN TCP allocations

seankhliao:
#### Description

Implements both client and server side support for TCP allocations and connections to peers.

Things to do:
- [ ] tests, what to test?
- [ ] add example(s)
- [ ] review/break out use of syscalls for port reuse
- [ ] review double lookup `internal/server/turn.go`

#### Reference issue

Fixes #143, #118

Issue #143, opened 2020-06-12
RFC6062 implementation


seankhliao:
I've read #118 and the associated branch [`rfc-6062-client`](https://github.com/pion/turn/tree/rfc-6062-client) and decided starting from scratch would be a better for me.

Currently on branch [`rfc6062`](https://github.com/pion/turn/tree/rfc6062) (sorry for confusing naming)  is a partially done modification of the client. Creating outbound requests `Connect` works as well as well as associating data connections `ConnectionBind`. Tested manually with `coturn`.

## Changes
`internal/proto`:
- add `ProtoTCP` 

`turn`:
- add `ProtoTCP` and `ProtoUDP`
- add `ConnectionID` type to main package
- add `ClientConfig.TransportProtocol`
- add `Client.transportProtocol`
- add `Client.nonce`
- add `Client.Connect()`
- add `Client.ConnectionBind()`

## Questions

- what kind of tests to add?
- successful authentication method should probably be stored (need to be reused)?
- `Client.Allocate` returns a `net.PacketConn`, not sure how that fits into TCP?
- the server will send `ConnectionAttempt` indications to the client for receiving connections, how should this be exposed to users? (also needs `CreatePermission`)
- handling refreshes

## example client

```go
package main

import (
	"fmt"
	"log"
	"net"

	"github.com/pion/turn/v2"
)

func main() {
	// Dial TURN Server
	turnServerAddr := "1.2.3.4:3478"
	conn, err := net.Dial("tcp", turnServerAddr)
	if err != nil {
		panic(err)
	}

	// Start a new TURN Client and wrap our net.Conn in a STUNConn
	// This allows us to simulate datagram based communication over a net.Conn
	cfg := &turn.ClientConfig{
		STUNServerAddr: turnServerAddr,
		TURNServerAddr: turnServerAddr,
		Conn:           turn.NewSTUNConn(conn),
		Username:       "user",
		Password:       "pass",
		TransportProtocol:       turn.ProtoTCP,
	}

	client, err := turn.NewClient(cfg)
	if err != nil {
		panic(err)
	}
	defer client.Close()

	// Start listening on the conn provided.
	err = client.Listen()
	if err != nil {
		panic(err)
	}

       // not sure what to do with relayConn
	relayConn, err := client.Allocate()
	if err != nil {
		panic(err)
	}
	defer func() {
		if closeErr := relayConn.Close(); closeErr != nil {
			panic(closeErr)
		}
	}()
[...]

Issue #118, opened 2020-01-20
Implement RFC-6062


kleky:
## Summary

Implement RFC-6062, Traversal Using Relays around NAT (TURN) Extensions for TCP Allocations

## Motivation

To provide a TCP relay client for use with a service such as Coturn. 

## Describe alternatives you've considered

Unable to find any clients that support RFC-6062.

## Additional context

Already discussed over Slack with a colleague of mine, **Lewis Jessett** and @Sean-Der : 

https://gophers.slack.com/archives/CAK2124AG/p1579206089185600

We've been given the clearance to contribute on this, but will probably need some guidance from the pion community.

Looking at the RFC I've tried to summarize what is required as follows:

- Client to setup a Turn ALLOCATE request for TCP ( REQUESTED-TRANSPORT attribute == 6)
- Refresh of ALLOCATION identical to UDP
- Ability for client to send a Connect request over control connection (including an XOR-PEER-ADDRESS attribute containing the transport address of the peer to which a connection is desired). -Upon success response the CONNECTION-ID attribute is returned to client.
- Ability for client to initiate another TCP connection using a different local transport address, then a ConnectionBind request over the new connection that includes the CONNECTION-ID attribute. Upon success, this is the client data connection.

Does a new data connection get created when a peer tries to connect, as per RFC-6062 Section 4.4?

Sean-Der:
Fantastic news! So excited you are able to use Pion @kleky I will make sure to help with the whole process/get things landed :)

I am reading the RFC now and will come back with my thoughts!

kleky:
Looks like we need to update vnet as it doesn't support TCP.

https://github.com/pion/transport/issues/27

Sean-Der:
@kleky 

Don't worry about adding `TCP` support to vnet. I don't think that blocks us getting this in at all. It is useful for testing how things behave in different network topologies, but I think it is less useful for TCP.

I read the RFC and seems *pretty* straight forward. We need to add some new handlers in [turn.go](https://github.com/pion/turn/blob/master/internal/server/turn.go) to handle `Connect` + `ConnectionBind` 

Then we need to make a `ConnHandler` equivalent [here](https://github.com/pion/turn/blob/master/internal/allocation/allocation.go#L216) this is the relay.

I am going to setup `coturn` and do a wireshark of TCP traffic. If you already have it setup would you mind sharing a `pcap` I can move a lot faster that way, but if you don't have one it is totally fine :)

kleky:
[...]

Issue #308, opened 2023-02-25
Provide easier way to create a TURN client


stv0g:
There has been a discussion initiated on whether we might want to introduce an easier API for creating a TURN client in #276:

There is a slightly related PR in https://github.com/pion/stun/pull/134, which introduces a new `stun.DialURI()` function.
It can be used in conjunction with `pion/turn.NewClient()`.

I am still pondering whether we also want to have a function in pion/turn to create a client directly by an TURN URI.

> I think this would be a nice addition that would make the life of people that just want a quick TURN client connection a lot more easier. Currently it takes some 70 lines of code in the UDP example to create the socket, set up the Client, and then calling client.Listen and client.Allocate, it would ne a big win to cut this down to 1-5 lines for simple use cases that do not require a fully fledged client.
>
> I'm wondering though how to pass the realm and the TURN credentials in to the Dial method, maybe we would need a simple Dialer that can hold these for the Dial?

I've introduced a new type in `pion/stun` to pass additional settings to the `stun.DialURI()` function:

```go
// DialConfig is used to pass configuration to DialURI()
type DialConfig struct {
	DTLSConfig dtls.Config
	TLSConfig  tls.Config

	Net transport.Net
}
```

We could do the same for `pion/turn`.
And thats where we probably need to discuss whether this addition would be a breaking API change as there exists already a `ClientConfig` struct.

Maybe it could look something like this:

```go
type ClientConfig struct {
	stun.DialConfig
	
	Username       string
	Password       string
	Realm          string
	Software       string
	RTO            time.Duration
	Conn           net.PacketConn // Listening socket (net.PacketConn)
	LoggerFactory  logging.LoggerFactory
}

// Connect by conn and STUN/TURN server addresses
func NewClientFromConn(conn net.PacketConn, stunAddr net.Addr, turnAddr net.Addr, cfg *ClientConfig) (*Client, error) { ... }

// Connect by URI
func NewClientFromURI(uri stun.URI, *cfg ClientConfig) (*Client, error) { ... }
```

Ideas for a better API are welcome :)

stv0g:
@rg0now I moved the discussion from #276 to a new issue here.

rg0now:
[...]

PR #315, opened 2023-03-22
RFC6062: Implement server side for TCP allocations

AndyLc:
#### Description
Builds off of https://github.com/pion/turn/pull/148

Currently the PR includes a commit for the client side too, but my plan is to rebase off https://github.com/pion/turn/pull/311 once it is merged. I am opening this PR for any initial feedback/review.

Main changes include:
- CreateAllocation now takes in a Protocol, either TCP or UDP, which describes the allocation <-> peer connection.
- allocation.NewManager now takes in a Protocol, either UDP/TCP/TLS which describes the client <-> turn connection type, not the allocation.
- Add AllocateListener, (I don't think we need AllocateConn anymore?), connectionHandler, connection maps, etc to handle TCP-connection based allocation logic.
- Move stun_conn.go to utils/stun_conn.go and add .Conn() method to get underlying net.Conn. This is because in handleConnectionBindRequest, after we receive the request and convert the turn<->peer connection to a data connection, we need to retrieve the original TCP connection to io.Copy to. STUNConn's ReadFrom does not seem appropriate.

Usage:
Start turn server (turn/examples/turn-server/tcp)
./tcp --public-ip 127.0.0.1 --port 3478 --users andy=secret

Run tcp client, (turn/examples/turn-client/tcp)
./tcp --host 127.0.0.1 --user andy=secret -server

Run another tcp client, which gets the relay IP from the other client and sends its own.
./tcp --host 127.0.0.1 --user andy=secret

Both clients should read and write a single message.

#### Reference issue
https://github.com/pion/turn/issues/143, https://github.com/pion/turn/issues/118


rg0now:
Thanks a lot, this is a super-useful work!

One suggestion: is it possible to rebase this PR to on top of #311? At least for me it'd be easier to see the client and the server side changes separately (I'm much more familiar with the server-side code).

> Add AllocateListener, (I don't think we need AllocateConn anymore?), connectionHandler, connection maps, etc to handle TCP-connection based allocation logic.

You mean the ``RelayAddressGenerator`` interface? Currently `Allocate` has the following signature:

```
AllocateConn(network string, requestedPort int) (net.Conn, net.Addr, error)
```

You suggest to rename this to `AllocateListener`, no? 

Now, I think we again face the same problem as on the client side: the thingie returned from `AllocateListener` should both be able to make new connections (implement our `transport.Dialer` interface) and accept new connections (implement `net.Listener`). So supposing that we create the new `Relay` interface that's both a `Dialer` and a `Listener` in `internal/transport` the new signature for `AllocateListener` would be something like the below:

```
AllocateListener(network string, requestedPort int) (transport.Relay, net.Addr, error)
```

But then we could also call this `AllocateRelay` no? Wdyt?

AndyLc:
Hi @stv0g and @rg0now,

I think this PR should now be ready for review now that the client side is merged. 

I believe we want the API to support a custom Dial (when an allocation receives a Connect request and creates a connection to peer via Dial), and a custom Listen (when an Allocation is created and listens to create data connection).

The custom listen is already possible through specifying a RelayAddressGenerator's `AllocateListener` that I added. What do you guys think is suitable for the custom Dial?

Perhaps a transport.Dialer can be specified in `turn.ServerConfig`?

client_test.go
```diff
@@ -227,7 +228,7 @@ func TestTCPClient(t *testing.T) {
 	require.NoError(t, err)
 
 	client, err := NewClient(&ClientConfig{
-		Conn:           NewSTUNConn(conn),
```
stv0g, line 230: Isn't this breaking the API by removing `NewSTUNConn`?

I agree that isnt a nice API. And I would much like to see it moved to `pion/stun` (see https://github.com/pion/turn/issues/308 for a related issue).

I think we currently have a lot STUN related code in `pion/turn`. E.g. `PerformTransaction` which would be better located in `pion/stun`. However, this is a different issue which we should not address in this PR.

server_config.go
```diff
@@ -67,7 +70,7 @@ func (c *PacketConnConfig) validate() error {
 // ListenerConfig is a single net.Listener to accept connections on. This will be used for TCP, TLS and DTLS listeners
 type ListenerConfig struct {
 	Listener net.Listener
-
+	Protocol string // TCP, or TLS
```
stv0g, line 73: Please indicate the expected string constants. E.g. "tcp", "tls" in the comment.

internal/server/turn_test.go
```diff
@@ -111,3 +112,67 @@ func TestAllocationLifeTime(t *testing.T) {
 		assert.Nil(t, r.AllocationManager.GetAllocation(fiveTuple))
 	})
 }
+
+func TestTCPAllocation(t *testing.T) {
+	t.Run("TCPAlloc", func(t *testing.T) {
```
stv0g, line 117: Do we really need a sub test here if there is only one in `TestTCPAllocation`?

internal/server/errors.go
```diff
@@ -19,7 +19,14 @@ var (
 	errFailedToCreateChannelData              = errors.New("failed to create channel data from packet")
 	errRelayAlreadyAllocatedForFiveTuple      = errors.New("relay already allocated for 5-TUPLE")
 	errUnsupportedTransportProtocol           = errors.New("RequestedTransport must be UDP or TCP")
+	errUnsupportedClientProtocol              = errors.New("cannot allocate tcp if client connection transport is not TCP/TLS")
```
stv0g, line 22: Please change lowercase `tcp` to `TCP`

stv0g: Also, I prefer to use the following wording for errors: `failed to ...`. E.g. in this instance `failed to allocate TCP...`

examples/turn-server/tls/main.go
```diff
@@ -77,6 +77,7 @@ func main() {
 					RelayAddress: net.ParseIP(*publicIP),
 					Address:      "0.0.0.0",
 				},
+				Protocol: "tls",
```
stv0g, line 80: Do we break this for existing users if they dont set the protocol?

internal/allocation/allocation_manager.go
```diff
@@ -103,15 +116,24 @@ func (m *Manager) CreateAllocation(fiveTuple *FiveTuple, turnSocket net.PacketCo
...
-	}
+	if a.RequestedTransportProtocol == proto.ProtoTCP {
+		listener, relayAddr, err := m.allocateListener("tcp", requestedPort)
+		if err != nil {
+			return nil, err
+		}
+		a.RelayListener = listener
+		a.RelayAddr = relayAddr
+	} else {
+		conn, relayAddr, err := m.allocatePacketConn("udp4", requestedPort)
```
stv0g, line 129: Not a change you introduced, but I am wondering why we are listening on IPv4 only..
Seems like we are not supporting `REQUESTED-ADDRESS-FAMILY` at all..

Another topic for another PR :D

internal/allocation/allocation_manager_test.go
```diff
@@ -187,7 +216,21 @@ func newTestManager() (*Manager, error) {
 
 			return conn, conn.LocalAddr(), nil
 		},
-		AllocateConn: func(network string, requestedPort int) (net.Conn, net.Addr, error) { return nil, nil, nil },
+		AllocateListener: func(network string, requestedPort int) (net.Listener, net.Addr, error) {
+			config := &net.ListenConfig{Control: func(network, address string, c syscall.RawConn) error {
+				var err error
+				c.Control(func(fd uintptr) {
```
stv0g, line 222: Check error of `c.Control()`

relay_address_generator_none.go
```diff
@@ -48,6 +51,23 @@ func (r *RelayAddressGeneratorNone) AllocatePacketConn(network string, requested
 	return conn, conn.LocalAddr(), nil
 }
 
+// AllocateListener generates a new Listener to receive traffic on and the IP/Port to populate the allocation response with
+func (r *RelayAddressGeneratorNone) AllocateListener(network string, requestedPort int) (net.Listener, net.Addr, error) {
+	config := &net.ListenConfig{Control: func(network, address string, c syscall.RawConn) error {
+		var err error
+		c.Control(func(fd uintptr) {
+			err = syscall.SetsockoptInt(int(fd), syscall.SOL_SOCKET, unix.SO_REUSEADDR|unix.SO_REUSEPORT, 1)
```
stv0g, line 59: Can you add a comment explaining the reasoning behind `SO_REUSEADDR/PORT`?

relay_address_generator_range.go
```diff
@@ -102,6 +106,49 @@ func (r *RelayAddressGeneratorPortRange) AllocatePacketConn(network string, requ
 	return nil, nil, errMaxRetriesExceeded
 }
 
+// AllocatePacketConn generates a new PacketConn to receive traffic on and the IP/Port to populate the allocation response with
```
stv0g, line 109: The comment does not match the function name

stv0g (changes requested):
Hi @AndyLc,

great work :-) I am a big fan of your PR like I was already for the client-side 👍 

I started reviewing it. But, will need to do this in several passed as its to big to review it at once..

Maybe @rg0now also has some capacity to support the reviewing process?

Thanks already!

Sean-Der:
This has been merged! It isn't available in a tagged release yet, but you can run the commands in `Usage` in master and it works :)

Thank you so much everyone for your time reviewing/talking about the feature. I am sorry I wasn't involved until recently. 

I would love to get peoples review on the API. We are going to do a major release of `pion/turn` in ~2 weeks after I get all the PRs/Issues triaged. 

Since we had to do a API breakage for this feature we should take this opportunity to fix anything else.

Issue #324, opened 2023-04-20
Add AllocationHandler


stv0g:
We should think about adding support for an `AllocationHandler` in a similar fashion like we already have for the `PermissionHandler`.

My motiviation behind it is the following:

We could use such a `AllocationHandler` to redirect clients via a `ALTERNATE-SERVER` response to other TURN servers.
However, we should give some flexibility to the user when and which alternate servers are signaled to the client.
This allows the user to define its own schemes to detect an overloaded server, or cycle through different available alternate servers.

rg0now:
Seems like a useful thing to add. Can you be a bit more specific though? What's the API? Where would we call the allocation handler, inside `server.CreateAllocation`? What would our allocation handler receive as argument and what to return?

stv0g:
> Where would we call the allocation handler, inside server.CreateAllocation?

This would be in [`internal/server.Server.handleAllocate()`](https://github.com/pion/turn/blob/f880e55089ad0e0cdb53c5a0e6b89954512dada6/internal/server/turn.go#L112).
Actually, the function already contains a comment:

https://github.com/pion/turn/blob/f880e55089ad0e0cdb53c5a0e6b89954512dada6/internal/server/turn.go#L112-L122

> What would our allocation handler receive as argument and what to return?

Good question. What about the following (just a first proposal):

```go
// AllocationHandler is a callback to ...
type AllocationHandler func(clientAddr net.Addr) (alternateServer *net.UDPAddr, stun.ErrorCode)
```

rg0now:
Do you also want to handle the allocation quota case (as per point 7 above) in the same `AllocationHandler` callback, or it is intended only for signaling the `ALTERNATE-SERVER` response?

Why I'm asking this is that every now and then our users complain about the lack of allocation quotas. So far, I failed to find sane API to support this: should we limit allocations per source IP? Or per TURN username/credential? Total allocations?

I guess [coturn implements the first approach and the third one](https://matrix-org.github.io/synapse/latest/setup/turn/coturn.html):

```
# consider whether you want to limit the quota of relayed streams per user (or total) to avoid risk of DoS.
user-quota=12 # 4 streams per video call, so 12 streams = 3 simultaneous relayed calls per user.
total-quota=1200
```

If we follow this course then I think your `AllocationHandler` API could also be used to support user quotas and total quotas. Wdyt?

stv0g:
Hi @rg0now,

[...]

PR #325, opened 2023-04-24
Add a callback to handle allocation requests

rg0now:
First take at #324.

The idea is to make it possible to customize the behavior of the TURN server in order to admit/deny allocation requests (e.g., to implement user quotas) or redirect clients to an alternate server. 

```go
type AllocationHandler func(clientAddr net.Addr) (alternateServer net.Addr, errorCode stun.ErrorCode)
```

Currently the allocation handler is per-listener (i.e., per `PacketConnConfig` and per `ListenerConfig`) and it is called just before the server would create a new allocation.

The PR contains a minimalistic implementation plus a test to get a feeling of the new API.

Some observations:
- client API: we will need to find a way to return the alternate server IP from `client.Allocate()`, should the server respond with a "Try Alternate" error (maybe the returned error could have a `StatusCode` and a `Parameter` of type `any` to return meaningful errors?) -- an alternative would be to automatically follow redirects in the client until we get to a usable server, but I guess this behavior should be made configurable,
- handle the `ALTERNATE-DOMAIN` attribute: the current API allows to return an alternate server address, but no alternate domain (see [here](https://www.rfc-editor.org/rfc/rfc8489.html#page-46)).
- user quotas: the current API is not enough to implement user quotas, in that we can already prevent over the quota allocations from being made but, given that we are only called on allocation creation but not at deletion, we cannot know when an allocation goes away, effectively resulting in blocking all access for the user *indefinitely*.

My takeaway is that most probably we'll need two distinct APIs to handle user quotas and redirects:
- the allocation handler will be fine for redirects (but maybe we can change the name to `RedirectHandler`, plus we may also need to handle the alternate domain case)
- user quotas need to track the lifecycle of allocations and  that will require a new callback, say, `QuotaHandler`:
  ```go
  type QuotaHandler func(clientAddr net.Addr, username string, new bool) (bool, error)
   ```
  setting `new=true` would mean it's a new allocation (otherwise we're deleting one), and returning (`false`, `nil`) would mean to deny access for the new allocation.


rg0now:
Sure, it's perfectly doable. Any thoughts on the client side API? How should we return the alt-server and alt-domain from `client.Allocate`? Or should we just merge this in ASAP and find out client-side API later? 

I'll update the redirect handler to return the alt-domain too ASAP.

stv0g:
Hi @rg0now,

> Any thoughts on the client side API? Or should we just merge this in ASAP and find out client-side API later?

I think, I would merge the server side handlers first and care about the client side changes via another PR.

> How should we return the alt-server and alt-domain from `client.Allocate`?

I think this would break the API which I would like to avoid.
I think we could easily handle this by returning a custom error type:

```go
type TryAlternateError struct {
    AlternateServer net.Addr
    AlternateDomain string
}
```

```go
func main() {
   ...

   conn, err := c.Allocate()
   if err != nil {
       var altErr *turn.TryAlternateError
       if errors.As(err, &altErr) {
           connectAlternate(altErr.AlternateServer, altErr.AlternateDomain)
       }
   }
}
```

stv0g:
On a second thought:

I think we should probably also simplify the signature of the allocation handler to:

```go
type AllocationHandler func(clientAddr net.Addr) error
```

Then we can use the same `TryAlternateError` from my previous comment.

stv0g:
I would also prefer to keep handlers closely related to the associated TURN methods.
E.g. I think a `AllocateHandler` can cover more possible use cases than a `RedirectHandler` or `QuotaHandler`.
However, you have a valid point regarding the release of allocated resources..

Following my line, we might also need to add an `AllocationExpiredHandler` which the user would need to hook for releasing resources.

stv0g:
In my projects, I usually try to use interfaces for the implementation of handlers:
But this is a larger, API-breaking change, which I am not sure we really need to do now:

```go
type AllocationHandler interface {
    Allocate(clientAddr net.Addr, username string) error
}

type AllocationExpiredHandler interface {
    AllocationExpired(clientAddr net.Addr, username string) error
}

type PermissionHandler interface {
    CreatePermission(clientAddr net.Addr, peerIP net.IP) error
    ExpirePermission(clientAddr net.Addr, peerIP net.IP) error
}

type AuthenticationHandler interface {
    Authenticate(username, realm string, srcAddr net.Addr) (key []byte, err error)
}

type ServerConfig struct {
    Handler any // this handler can implement any of the previous interfaces
}
```

The advantage of tying the handlers to a struct would be, that the user can easily associate state to the handlers (like a quota database).  Or implement handlers which implement subset of the aforementioned interfaces.

rg0now:
This makes a lot of sense. In Kubernetes we use API versions to handle breaking API changes: we release a new version (say, v1alpha2 on top of v1alpha1), an automatic converter that upgrades an API call from the old version, and then we deprecate the old API and stop supporting it after 2-3 release. I really miss API versioning from Go (or any other programming language for that matter): it would make such API upgrades much simpler. 

Anyway, here's the current plan, let me know if you're fine with it:
- add a new handler to `PacketConnConfig` and `ListenerConfig` types with the signature:
  ```
  type AllocationHandler func(clientAddr net.Addr) error
  ```
- insert a call to the allocation handler into `internal/server/turn.go:handleAllocateRequest`
- add an example that demonstrates a TURN server that returns a redirect error for allocation requests
- worry about the client-side interface later.

As per the last point, I'm happy to implement the `TryAlternateError` API too, let me know if I should do it now or put it into another PR.

stv0g:
Hi @rg0now,

I am missing just one thing: the `AllocateHandler`s signature is not very feature proof. We cant handle additional attributes if we want to use the handler for other purposes (e.g. quota handling).

I am thinking about attributes like:
- `USERNAME`
- `REQUESTED-TRANSPORT` (for #311, #315)
- `REQUESTED-ADDRESS-FAMILY`
- `ADDITIONAL-ADDRESS-FAMILY`
- ...

What do you think about simply passing the `stun.Message` as a second parameter?

stv0g:
> What if we create a struct with all the data the caller would ever need to make a decision (like the username, transport, etc.) and pass that to the callback?

I had the same idea. I think this would work as well as we can extend the structure in the future 👍

rg0now:
> The problem I see here is that we do not know the five-tuple at the time when `CreateAllocation` handler is called. Hence, the quota handler can not correlate the quota allocation (`CreateAllocation`) to the release (`AllocationDeleted`).

Hmm, I really want to get this right so let's see what is that I don't understand.

Let's pass a struct like this to the `CreateAllocation` handler:

```go
type AllocationDescriptor struct {
	Username                   stun.Username
	FiveTuple                  allocation.FiveTuple
	RequestedTransportProtocol allocation.Protocol
	RequestedAddressFamily     proto.RequestedAddressFamily
	AdditionalAddressFamily    proto.RequestedAddressFamily
}
```

The allocation handler interface would look like this:

```go
type AllocationHandler interface {
	// CreateAllocation is called before the server making a new allocation. Return an error to prevent the allocation from being made.
	CreateAllocation(desc AllocationDescriptor) error
	// AllocationDeleted is called by the server after deleting an allocation. This is useful to track the lifetime of allocations.
	AllocationDeleted(fiveTuple *allocation.FiveTuple)
}
```

We would call the `CreateAllocation` handler in `internal/server/turn.go:handleAllocateRequest`. The structure of this longish function would look like this:

```go
func handleAllocateRequest(r Request, m *stun.Message) error {
	// integrity check

	// create client-side 5-tuple
	fiveTuple := &allocation.FiveTuple{
		SrcAddr:  r.SrcAddr,
		DstAddr:  r.Conn.LocalAddr(),
		Protocol: allocation.UDP,
	}

	// check if allocation already exists
	if alloc := r.AllocationManager.GetAllocation(fiveTuple); alloc != nil { ... }

	// handle RequestedTrasnport, AttrDontFragment, ReservationToken and EvenPort

	// call the allocation handler's CreateAllocation callback
	desc := alloc.NewAllocationDescriptor()
	err := r.AllocationManager.allocationHandler.CreateAllocation(desc)
	if err != nil {
		// return err to client
	}

	// create the allocation...
}
```

Another option here would be to call the `CreateAllocation` handler from `internal/allocation/allocation_manager.go:CreateAllocation`, this seems more "self-contained" but at this point we lost our `stun.Message` so we have to pass the necessary `AllocationDescriptor` fields (username, etc.) as arguments.

Finally, `internal/allocation/allocation_manager.go:DeleteAllocation` would look like the below

```go
func (m *Manager) DeleteAllocation(fiveTuple *FiveTuple) {
	m.allocationHandler.AllocationDeleted(fiveTuple)

	// delete and close allocation
}
```

Wdyt?

stv0g:
Hey @rg0now,

thanks for the detailed description :-) I think I got this wrong. I was under the impression that the five-tuple contains:

- Client IP+Port
- Relayed IP+Port
- Protocol

However, you seem to be correctly using the clients connection local address (from server perspective) instead of the relayed address.

I still need to understand why this works. This way, it seems like the client can not create multiple allocations using the same connection to the server as these would all use the same five-tuple?

rg0now:
> I still need to understand why this works. This way, it seems like the client can not create multiple allocations using the same connection to the server as these would all use the same five-tuple?

Yes, this strikes me as odd too, but it seems that each `turn.Client` can support only a single allocation (the client actually caches the `relayedConn`). This seems like an assumption we don't quite document. If you want more allocations, create more `Client`s, each with a separate `net.PacketConn` and, hence, with a separate `FiveTuple`.

The useful consequence of this assumption is that we can use the `FiveTuple` to identify allocations, like in `internal/allocation/allocation_manager.go:GetAllocation`:

```go
	m.lock.RLock()
	defer m.lock.RUnlock()
	return m.allocations[fiveTuple.Fingerprint()]
}
```

Maybe the reason of why you're finding this confusing now is that you're deep into the implementation of RFC6062 TCP allocations and the client model is so much different for TCP.

stv0g:
Okay, I've read the TURN RFC (again) 😁

It clearly says that the five-tuple is identifying the client-server connection and that a client may only have a single or dual allocation active (dual as in dual-stack when requesting an allocation with relayed addresses for multiple address familiies e.g. IPv4 + IPv6).

See: https://datatracker.ietf.org/doc/html/rfc8656#name-allocations-2

This now also resolves some of my confusion in regard to the TCP allocation PRs which we are currently working on 😃 

So now, I can fully support your proposed implementation 👍 

Just with one remark:

I would try to keep the handlers out of the `AllocationManager` and just to `Server` since I think it would be cleaner if the  `AllocationManager` does not need to care about the handlers. I would try to keep them just to the `Server` types.

I still try to understand why somebody thought it was a good idea to have both a public `Server` and internal `Server` type -.- I believe, this makes a lot of stuff more difficult to understand. But thats a different topic :D

Sean-Der:
@rg0now Anything from this PR you would like to keep? Is all this covered by EventHandlers now?

rg0now:
Implemented in #419

PR #340, opened 2023-06-27
Partial revert of 79fbb4e3

### LINKED ISSUE #339: github.com/pion/ice/v2@v2.3.8/gather.go:657:21: cannot use turnServerAddr (variable of type string) as net.Addr value in struct literal: string does not implement net.Addr (missing method Network)
github.com/pion/ice/v2@v2.3.8/gather.go:657:21: cannot use turnServerAddr (variable of type string) as net.Addr value in struct literal: string does not implement net.Addr (missing method Network)

Sean-Der:
This commit broke public API, this reverts those changes

Resolves #339

### DIFF
--- a/client.go
+++ b/client.go
@@ -37,8 +37,8 @@ const (
 
 // ClientConfig is a bag of config parameters for Client.
 type ClientConfig struct {
-	STUNServerAddr net.Addr // STUN server address
-	TURNServerAddr net.Addr // TURN server address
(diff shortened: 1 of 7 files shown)


Sean-Der:
@stv0g when you get a chance mind reviewing this? I don't think I broke anything, but didn't want to leave users in a bad state.

thanks!

PR #343, opened 2023-07-05
Move GenerateAuthKey to SHA256

### LINKED ISSUE #342: Consider replacing MD5
The GenerateAuthKey function [here](https://github.com/pion/turn/blob/83e705812565d5f7e7ffb19aa5eee85d4f117960/server_config.go#L97) uses MD5, which [is](https://en.wikipedia.org/wiki/MD5#Security) [insecure](https://security.stackexchange.com/a/19908).

Consider using a modern hashing algorithm like SHA256 if possible.

okyanusoz:
#### Description

This PR modifies GenerateAuthKey to use SHA256 instead of MD5 to avoid possible collisions. See #342 for more details.

#### Reference issue
Fixes #342

### DIFF
--- a/server_config.go
+++ b/server_config.go
@@ -4,7 +4,7 @@
 package turn
 
 import (
-	"crypto/md5" //nolint:gosec,gci
+	"crypto/sha256" //nolint:gosec,gci
 	"fmt"
 	"net"
 	"strings"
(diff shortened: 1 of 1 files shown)


jech:
We cannot do that, the hashing function is specified in Section 15.4 of RFC 5389. However, the use of MD5 is not a security issue here, for at least two reasons:
  - in RFC 5389 usage, the password is computed using HMAC-SHA1, and the MD5 hash only serves to provide a fixed-length key from variable-length data, not for security purposes;
  - MD5 is vulnerable to collision attacks, not to reverse image attacks, so this particular usage is fine.

PR #345, opened 2023-07-22
Prevent nil pointer dereference

boks1971:
#### Description
Check for stunServerAddr != nil before dereferencing it.


davidzhao (approved):
lg

PR #356, opened 2023-09-18
Fix out of bounds access panic

boks1971:
Drop packet exceeding size of input buffer.


boks1971:
@reviewers: I put in a `continue` to drop the bad packet. Instead, should we send in the truncated packet to the handler? (after setting proper bounds).

cnderrauber (approved):

PR #357, opened 2023-09-21
Use multi-stage docker build for examples/turn-server/simple to reduce container image size

levaitamas:
The current docker build of [examples/turn-server/simple](https://github.com/pion/turn/tree/master/examples/turn-server/simple) results a large (944MB) image.
This PR implements [multi-stage build](https://docs.docker.com/build/building/multi-stage/) that reduces image size to 10.2MB.

### DIFF
--- a/examples/turn-server/simple/Dockerfile
+++ b/examples/turn-server/simple/Dockerfile
@@ -1,7 +1,28 @@
 # SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
 # SPDX-License-Identifier: MIT
 
-FROM golang:latest
+FROM golang:alpine as builder
+
+ARG VERSION=master
+
+RUN apk add --no-cache git
+
+WORKDIR /build
+# Clone Source using GIT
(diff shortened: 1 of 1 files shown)


examples/turn-server/simple/Dockerfile
```diff
@@ -1,7 +1,28 @@
 # SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
 # SPDX-License-Identifier: MIT
 
-FROM golang:latest
+##### builder
```
stv0g, line 4: I think this comment is superfluous. The `AS builder` should be sufficient

stv0g:
```suggestion
```

stv0g (approved):
Looks great! Would you mind removing the the comment and merge it?

examples/turn-server/simple/Dockerfile
levaitamas: Fixed in [3f83860](https://github.com/pion/turn/pull/357/commits/3f83860de7001522e614951e959766dba7db37bf)

levaitamas:
Thanks @stv0g!

PR #359, opened 2023-10-23
Use crypto/rand instead of math/rand

dkess:
math/rand isn't suitable for security-sensitive contexts. This PR replaces it with crypto/rand (which uses the somewhat inconvenient big.Int type).

### DIFF
--- a/internal/server/turn.go
+++ b/internal/server/turn.go
@@ -7,11 +7,14 @@ import (
 	"fmt"
 	"net"
 
+	"github.com/pion/randutil"
 	"github.com/pion/stun/v2"
 	"github.com/pion/turn/v3/internal/allocation"
 	"github.com/pion/turn/v3/internal/ipnet"
 	"github.com/pion/turn/v3/internal/proto"
 )
+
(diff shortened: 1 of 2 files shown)


Sean-Der:
Nice catch! Thank you @dkess 

MInd using [randutil.GenerateCryptoRandomString](https://github.com/pion/randutil/blob/master/crypto.go#L13C6-L13C32)

After pion/turn was created we started moving our duplicated random/crypto logic to this repo. Sorry to make you do the extra work :/

dkess:
No problem. Updated the PR. Thanks for the pointer.

Sean-Der (approved):

Sean-Der:
merged!

@dkess any interest in staying involved? I would love to add you to the Pion org. If you have other things that could be improved with this repo, would love to support you :)

dkess:
Happy to be part of the org :)

I don't know of any improvements yet, but I do plan to use the ICE library for an upcoming project at my workplace. I'll definitely be keeping an eye on the project and plan to participate as needed.

PR #370, opened 2024-02-04
lt_cred support for TURN REST format

giavac:
#### Description

This PR adds support for ephemeral credentials, as initially described in https://datatracker.ietf.org/doc/html/draft-uberti-behave-turn-rest-00 and widely used by https://github.com/coturn/coturn


giavac:
I see the failed checks, I'll amend.

examples/turn-server/lt-cred-turn-rest/main.go
```diff
@@ -0,0 +1,72 @@
...
+
+	// NewLongTermAuthHandler takes a pion.LeveledLogger. This allows you to intercept messages
+	// and process them yourself.
+	logger := logging.NewDefaultLeveledLoggerForScope("lt-creds", logging.LogLevelTrace, os.Stdout)
+
+	s, err := turn.NewServer(turn.ServerConfig{
+		Realm: *realm,
+		// Set AuthHandler callback
+		// This is called everytime a user tries to authenticate with the TURN server
+		// Return the key for that user, or false when no user is found
```
rg0now, line 50: Please remove useless comment. Also adding a short explanation might help people trying to adopt the feature.

examples/turn-server/lt-cred-turn-rest/main.go
```diff
@@ -0,0 +1,72 @@
+// Package main implements a TURN server using
```
rg0now, line 4: Please add copyright stanza as per the failing REUSE test. See `examples/turn-server/lt-cred/main.go` for an example.

examples/turn-server/lt-cred-turn-rest/main.go
```diff
@@ -0,0 +1,72 @@
...
+// ephemeral credentials.
```
rg0now, line 5: Please add a short description of what's going on here to `examples/README.md`.  No need to go into details, just the basic usage like [here](https://github.com/giavac/turn/tree/master/examples#lt-creds). This can go right after the existing `lt-creds` description.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/pion/turn/v3
 
-go 1.13
+go 1.19
```
rg0now, line 3: I'm not sure we are allowed to upgrade the required Go version here. That would be a major break and may require a major release (not sure about this). Currently we are at 1.13 and my understanding is that there's nothing in your code that absolutely requires a higher version I'd rather not bump the Go version here.

ps: not that I full understand the min Go version requirements across the entire pion project (some code like DTLS is still at 1.13 whole some like mediadevices require 1.19)...

lt_cred.go
```diff
@@ -57,3 +67,31 @@ func NewLongTermAuthHandler(sharedSecret string, l logging.LeveledLogger) AuthHa
 		return GenerateAuthKey(username, realm, password), true
 	}
 }
+
+// LongTermTURNRESTAuthHandler returns a turn.AuthAuthHandler used with Long Term (or Time Windowed) Credentials.
```
rg0now, line 71: Can you please add one line description of the main purpose of the REST API? I'm thinking of something along the following lines:

```
// LongTermTURNRESTAuthHandler returns a turn.AuthAuthHandler that can be used to authenticate
// time-windowed TURN credentials generated by the TURN REST API described in
// https://datatracker.ietf.org/doc/html/draft-uberti-behave-turn-rest-00
//
// The supported format of is timestamp:username, where username is an arbitrary user id and the
// timestamp specifies the expiry of the credential.
```

If you are at it: can you please also fix the URL in the doc of `NewLongTermAuthHandler`: the orignal URL  https://tools.ietf.org/search/rfc5389#section-10.2 does not seem to work anymore for some reason, plus RFC 5389 is now obsoleted by RFC 8489, so we should use this URL instead: https://datatracker.ietf.org/doc/html/rfc8489#section-9.2.

rg0now (changes requested):
Just a minor final push and it's in!..:-)

rg0now:
Dear @giavac, would you be so kind to collapse your commits into a single one? (Just `reset --soft <branch-root>` and then commit everything in one and force-push). It seems we will have to manually merge due to the inconclusive tests (I'm trying to ask around at Slack, still  no info) and I don't want the squash-merge to overwrite authorship info in your commits.

giavac:
Thanks @rg0now - it should be OK now?

giavac:
@rg0now maybe I could try to increase the code coverage, if that's what's holding the last check?

rg0now:
Unfortunately that won't fix this (codecov checks are not "required") but it would definitely help. The problem is the pending results for the Go 1.21 tests. Since I don't seem to receive a definite answer at Slack, I think we'll just merge manually anyway.

rg0now:
Thanks a lot @giavac, merged

giavac:
Thanks for the review and guidance @rg0now

PR #371, opened 2024-02-06
Use signed timestamp for nonce

### LINKED ISSUE #361: Memory leak because server.go nonces increasing
### Your environment.
  - Version: latest
  - Browser: any (server problem)

### What did you do?

In our production environment, TURN server runs a long time.

### What did you expect?
Not expect memory leak.

### What happened?
RSS memory of TURN server is increasing.

Troubleshoot:
I use Golang pprof to profile the server and find that `nonces` map of the global server is large. The code about nonce creation and deletion is in `authenticateRequest`. After a normal TURN request, the nonce remains in the map.

How to solve? I propose three methods:
1、WebRTC turn_server.cc ( https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/p2p/base/turn_server.cc;drc=8e78783dc1f7007bad46d657c9f332614e240fd8;l=385?q=turn_server.cc ) decodes nonce to get timestamp. So no need to save nonce and timestamp in a map.

2、coturn seems to save nonce to a session. You known, TCP (client to TURN) has connection, so we can save nonce to something like connection (in pion/turn `readListener` ). If the connection is broken, nonce also doesn't exist. I think UDP (client to TURN)  is quite troublesome, we must create a structure to represent the UDP session.

3、Using a timer to delete expired nonce in the nonces map. However nonce expiration time is one hour, the map may remain large.

I prefer the first method.

paulwe:
#### Description
implements nonce generation using the method from chromium (https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/p2p/base/turn_server.cc;l=385;drc=8e78783dc1f7007bad46d657c9f332614e240fd8?q=turn_server.cc). create a random hmac and use it to hash the timestamp of the authentication request. validate by checking the hash and time elapsed since the nonce was generated.

#### Reference issue
Fixes #361

### DIFF
--- a/internal/server/nonce.go
+++ b/internal/server/nonce.go
@@ -1,0 +1,71 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package server
+
+import (
+	"crypto/hmac"
+	"crypto/rand"
+	"crypto/sha256"
+	"encoding/binary"
+	"encoding/hex"
+	"fmt"
+	"time"
+)
+
+const (
+	nonceLifetime  = time.Hour // See: https://tools.ietf.org/html/rfc5766#section-4
+	nonceLength    = 40
+	nonceKeyLength = 64
+)
+
+// NewNonceHash creates a NonceHash
+func NewNonceHash() (*NonceHash, error) {
+	key := make([]byte, nonceKeyLength)
+	if _, err := rand.Read(key); err != nil {
+		return nil, err
+	}
+
+	return &NonceHash{key}, nil
+}
+
+// NonceHash is used to create and verify nonces
+type NonceHash struct {
+	key []byte
+}
+
+// Generate a nonce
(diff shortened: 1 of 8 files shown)


internal/server/nonce.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	return &NonceHash{key}, nil
+}
+
+type NonceHash struct {
+	key []byte
+}
+
+func (g *NonceHash) Generate() (string, error) {
```
boks1971, line 35: Not sure if Pion linter complains, maybe name the variable `n` to follow the convention of first letter of type?

boks1971 (approved):
LGTM!

rg0now:
Looks good to me too. 

One minor (and probably insignificant) point: the signature of `server.Request` changes, I'm wondering whether this is public API. I'd say no, but the line between what's public and what's not is really fuzzy in Go.

paulwe:
`server.Request` is an internal package so it can't be imported by other packages.

PR #382, opened 2024-04-04
Change listening port in test

Sean-Der:
127.0.0.133:54321 fails to bind on macOS


rg0now:
We've already been bitten by this: https://github.com/pion/turn/issues/293#issuecomment-1410586172 :-D

Sean-Der:
I wonder why macOS has this behavior, seems arbitrary! Luckily that wasn't too hard to debug :) 

The `mockClient` causing a crash on GitHub Actions (and not locally) was way more painful 😱

PR #383, opened 2024-04-04
Support stand alone STUN server mode

Sean-Der:
Running in stand alone STUN server does not start
due to packet conn config requiring relay address config during validation.

To enable that mode, bypass validation if relay address config in packet conn config is nil.

Note that it is still validated in `ListenerConfig`.


boks1971 (approved):
lgtm!!! Thank you @Sean-Der

PR #393, opened 2024-04-20
Prevent heap allocations in FiveTuple Fingerprint

paulwe:


davidzhao (approved):
nice, this will be a solid speedup

rg0now:
We've already tried to track this down with @stv0g, see #299 and #304. This seems like a much better solution and the immediate effect is indeed massive performance improvement in `GetAllocation`  based on my benchmarks (`go test -cpuprofile cpu_895851.prof -bench=. -run=XXX -benchtime=200000x`).

`GetAllocation` before the merge: `UDPAddr.String` dominates performance.
profile_895851

`GetAllocation` after the merge: `FiveTupeFingerprint` got supercheap.
profile_55407d

PR #402, opened 2024-07-16
Add callback for TURN authentication success

renandincer:
#### Description

Creates a new callback for users to be notified by successful TURN authentication. Together with the authHandler function, this can help determine the ratio of successful vs unsuccessful authentication attemps per username. This information then can be used to implement methods to prevent attacks to guess credentials similar to fail2ban.

There's been some discussion around implementing various usage quotas and limits within this package. An auth success callback can be used to implement some of these given that a succesful source IP address can be reliably linked to usernames. Together with maximum allocation time limit and permission timeouts per RFC, allocation refesh requests (requires auth) and configurable channel bind timeouts, fivetuple -> username usage can be reliably tracked outside of this package.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -28,6 +28,7 @@ type Request struct {
 
 	// User Configuration
 	AuthHandler        func(username string, realm string, srcAddr net.Addr) (key []byte, ok bool)
+	AuthSuccess        func(username string, realm string, srcAddr net.Addr)
 	Log                logging.LeveledLogger
 	Realm              string
 	ChannelBindTimeout time.Duration
--- a/internal/server/turn_test.go
+++ b/internal/server/turn_test.go
@@ -84,6 +84,8 @@ func TestAllocationLifeTime(t *testing.T) {
 		staticKey, err := nonceHash.Generate()
 		assert.NoError(t, err)
 
+		authSuccessCallbackTimes := 0
+
 		r := Request{
 			AllocationManager: allocationManager,
 			NonceHash:         nonceHash,
@@ -93,13 +95,16 @@ func TestAllocationLifeTime(t *testing.T) {
 			AuthHandler: func(string, string, net.Addr) (key []byte, ok bool) {
 				return []byte(staticKey), true
 			},
+
+			AuthSuccess: func(username string, realm string, srcAddr net.Addr) {
+				authSuccessCallbackTimes++
+			},
 		}
 
 		fiveTuple := &allocation.FiveTuple{SrcAddr: r.SrcAddr, DstAddr: r.Conn.LocalAddr(), Protocol: allocation.UDP}
 
 		_, err = r.AllocationManager.CreateAllocation(fiveTuple, r.Conn, 0, time.Hour)
 		assert.NoError(t, err)
-
 		assert.NotNil(t, r.AllocationManager.GetAllocation(fiveTuple))
 
 		m := &stun.Message{}
@@ -109,7 +114,12 @@ func TestAllocationLifeTime(t *testing.T) {
 		assert.NoError(t, (stun.Realm(staticKey)).AddTo(m))
 		assert.NoError(t, (stun.Username(staticKey)).AddTo(m))
 
+		assert.NoError(t, handleCreatePermissionRequest(r, m))
+		assert.Equal(t, 1, authSuccessCallbackTimes)
+
 		assert.NoError(t, handleRefreshRequest(r, m))
+		assert.Equal(t, 2, authSuccessCallbackTimes)
+
 		assert.Nil(t, r.AllocationManager.GetAllocation(fiveTuple))
 	})
 }
--- a/internal/server/util.go
+++ b/internal/server/util.go
@@ -97,6 +97,10 @@ func authenticateRequest(r Request, m *stun.Message, callingMethod stun.Method) 
 		return nil, false, buildAndSendErr(r.Conn, r.SrcAddr, err, badRequestMsg...)
 	}
 
+	if r.AuthSuccess != nil {
+		r.AuthSuccess(usernameAttr.String(), realmAttr.String(), r.SrcAddr)
+	}
+
 	return stun.MessageIntegrity(ourKey), true, nil
 }
 
--- a/server.go
+++ b/server.go
@@ -24,6 +24,7 @@ const (
 type Server struct {
 	log                logging.LeveledLogger
 	authHandler        AuthHandler
+	authSuccess        AuthCallback
 	realm              string
 	channelBindTimeout time.Duration
 	nonceHash          *server.NonceHash
@@ -60,6 +61,7 @@ func NewServer(config ServerConfig) (*Server, error) {
 	s := &Server{
 		log:                loggerFactory.NewLogger("turn"),
 		authHandler:        config.AuthHandler,
+		authSuccess:        config.AuthSuccess,
 		realm:              config.Realm,
 		channelBindTimeout: config.ChannelBindTimeout,
 		packetConnConfigs:  config.PacketConnConfigs,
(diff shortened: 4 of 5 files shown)


renandincer:
(forced pushed to resolve the linter issue)

Sean-Der:
Great call out @rg0now it would be great if full life cycle could be tracked! 

@rg0now since this isn't a breaking change feel free to merge/change however you see fit! I also am in support of you starting a `/v4` at anytime you see fit. You have done an exceptional job (better then me!) with maintaining pion/turn :)

Tell me what I can do to support you. I think you are a better advocate/understand what 'next-gen TURN' looks like.

renandincer:
Hey @rg0now, yes I reviewed those PRs, and wanted to keep the scope of this as small as possible due to the discussion there - keeping the ALTERNATE-SERVER use cases out of it. 

> You mention that the intention of the PR is to ensure that "fivetuple -> username usage can be reliably tracked". Can we handle this with the AuthHandler callback? I mean, the application provides the AuthHandler so they may also provide the fivetuple-tracking functionality as well in the AuthHandler, no?

AuthHandler will be called regardless of authentication success. This prevents a reliable mapping between fivetuple to usernames so that's why a callback only for success is desirable.

> Currently the AuthSuccess callback is called every time (1) an allocation is created or refreshed, or a permission or a channel is created. Is this the intended semantics?

This is the intended semantics. It keeps 1-1 mapping with AuthHandler, so it's simple and doesn't cause any breaking changes while achieving the two goals: (1) create a way to detect failed vs successful requests per srcAddr, and (2) create a way map fiveTuple to username.

> Shouldn't we rather put the callback site into the CreateAllocation function of the allocation.Manager? This may be perhaps semantically correct, in turn we lose the username and perhaps the realm at this point (but it may not be difficult to store these in turn.Request and pass in the full turn.Request into AllocationManager.CreateAllocation instead of only r.Conn). In other words, would you be better off with a AllocationCreated callback rather then an AuthSuccess callback?

I do fully agree with you. I think what you described is a superior way of doing this. Callback for AllocationCreated and AllocationDeleted along with username and fivetuple in the function signature like so:

`AllocationCreated(username string, fiveTuple FiveTuple)`
`AllocationDeleted(username string, fiveTuple FiveTuple)`

Somewhat related, getting the `AllocatePacketConn` function signature to accept username and fiveTuple will (or a way to access the net.Conn assigned by `AllocatePacketConn` when allocation is created) is also desirable to be able to implement rate limiting style functions on a per-username basis).

> Are you interested in handling user quotas? Would you be willing to add a AllocationDeleted callback too?

RFC does not have a explicit way of deleting a allocation or permission so timeouts are enough for keeping track of usage accurately along with this auth success callback function.

I'm willing to discuss other implementations, contributing AllocationDeleted/AllocatedCreated/etc. 

Yep. Let's get on a call and discuss :)

rg0now:
> wanted to keep the scope of this as small as possible due to the discussion there - keeping the ALTERNATE-DOMAIN style use cases out of it.

Full ack.

> AuthHandler will be called regardless of authentication success. This prevents a reliable mapping between fivetuple to usernames so that's why a callback only for success is desirable.
> This is the intended semantics. It keeps 1-1 mapping with AuthHandler, so it's simple and doesn't cause any breaking changes while achieving the two goals: (1) create a way to detect failed vs successful requests per srcAddr, and (2) create a way map fiveTuple to username.

I see. It seems worth supporting this use case.

> Somewhat related, getting the `AllocatePacketConn` function signature to accept username and fiveTuple will (or a way to access the `net.Conn` assigned by `AllocatePacketConn` when allocation is created) is also desirable to be able to implement rate limiting style functions on a per-username basis).

I think the way user quotas could be supported is through the `AuthHandler`: when a user has filled their quota we start rejecting new allocations from the `AuthHandler` where we have enough context to do so. But this workflow is somewhat unintuitive, that's why we should provide an example for supporting user quotas in the server. Wdyt?

> RFC does not have a explicit way of deleting a allocation or permission so timeouts are enough for keeping track of usage accurately along with this auth success callback function.

I'm not sure I get this. We (the user) need to be notified when an allocation goes away, otherwise we cannot keep track of user quotas.

> I'm willing to discuss other implementations, contributing AllocationDeleted/AllocatedCreated/etc.

Here is a *final* API I have in mind:

```
type EventContext struct {
	Username, Realm  string   // add to allocation.Allocation
	Network          string   // from allocation.FiveTuple
	SrcAddr, DstAddr net.Addr // from allocation.FiveTuple
}

type EventHandlers struct {
	AuthSuccess       func(context EventContext)                       // called from server.authenticateRequest
	AuthFailure       func(context EventContext)                       // called from server.authenticateRequest
	AllocationCreated func(context EventContext, requestedPort int)    // called from allocation.CreateAllocation
	AllocationRemoved func(context EventContext)                       // called from allocation.DeleteAllocation
	PermissionCreated func(context EventContext, ip net.IP)            // called from allocation.AddPermission
	PermissionRemoved func(context EventContext, ip net.IP)            // called from allocation.RemovePermission
	ChannelCreated    func(context EventContext, channelNumber uint16) // called from alloction.AddChannelBind (this also calles PermissionCreated)
	ChannelRemoved    func(context EventContext, channelNumber uint16) // called from alloction.RemoveChannelBind (this also calles PermissionRemoved)
}
```

This is deliberately not an interface so people will not have to implement all event handlers, just the ones they are interested in. Of course we don't have to implement all this: the first two callbacks will do for now (this will already handle your use case) and then we can work together to gradually implement the full API. Wdyt?

Let's chat! Wrote an email!...

renandincer:
> We (the user) need to be notified when an allocation goes away, otherwise we cannot keep track of user quotas.

With the knowledge of:
- Successful authentication
- Username
- IP:Port tuple for the source
- Various timeouts in TURN
You can keep track of a map between username and src ip:port tuple, which can drive user quotas.

The authSuccess function is called for every:
- Allocate Request (maximum allocation time 1 hour per RFC recommendation)
- Allocation Refresh Request - any extension to 1 hour RFC)
- Channel Bind Request (configurable timeout - 10 minutes by default)
- Create Permission Request (permission timeout 5 minutes per RFC requirement)

Since one can't send or receive packets without channel bind or permission or authenticated packet, you can assume ip:port tuple to username mapping to be 10 minutes by default in this situation.

Obviously this is a unintuitive way of implementing this.

renandincer:
> I think the way user quotas could be supported is through the AuthHandler: when a user has filled their quota we start rejecting new allocations from the AuthHandler where we have enough context to do so. But this workflow is somewhat unintuitive, that's why we should provide an example for supporting user quotas in the server. Wdyt?

I think example is a good idea in general whatever we implement. :)

> Here is a final API I have in mind:

I like this API, but I would like to sleep on it a little. I am not sure if ChannelCreated/ChannelRemoved callbacks are very useful. Do you think there is a good use for them?

renandincer:
> Problem is: any way to solve this seems to require a breaking change. My suggestion would be to get the lifecycle event handlers straight and once we're happy with it only then break the API.

One option could be to provide an API to get a pointer to the net.Conn given username (or something from the allocation)

rg0now:
> I like this API, but I would like to sleep on it a little. I am not sure if ChannelCreated/ChannelRemoved callbacks are very useful. Do you think there is a good use for them?

One use case that I have in mind is making eBPF acceleration less intrusive (see https://github.com/pion/turn/pull/360). The current PR hooks directly into `AddChannelBind` and `DeleteChannelBind` to create/remove the eBPF offload mappings for the channel. Should we have the lifecycle event handlers, we could implement the eBPF acceleration completely outside the base pion/turn package without having to distribute the eBPF binary object files with the source code (something nobody seems to like for understandable reasons).

renandincer:
To close the loop here for those watching, we had a very good call.

I agree that a interface similar to what @rg0now suggested is the way forward. This way, in the future, extra information from the allocation can be exported with relative ease without having to make breaking changes to the API.

I will close this PR and follow up with a implementation in coming months unless somebody else takes on before then :)

rg0now:
Hi @renandincer any progress on this? I plan to take a stride at an implementation but I don't want to do duplicate work...

PR #419, opened 2024-11-07
Add callback API to track allocation lifeycle

rg0now:
Fixes https://github.com/pion/turn/issues/324, https://github.com/pion/turn/pull/325, and https://github.com/pion/turn/pull/402

This PR adds a set of callbacks that can be used to track the lifecycle of TURN allocations in the server. This allows user code to be called whenever an authentication request has been processed (this can be used to mitigate DoS attacks against a server, see https://github.com/pion/turn/pull/402), when an allocation has been created/removed (e.g., to implement user tracking for adding quota support, see https://github.com/pion/turn/issues/324), when a permission has been created/removed (e.g., to log the peers a user tries to reach via the server), a channel has been created/removed (this will simplify integrating external TURN accelerators, see https://github.com/pion/turn/pull/360), or when an allocation ends with an error.

The implementation adds a new `EventHandlers` struct to the `ServerConfig`, defined as follows:

```
type EventHandlers struct {
  OnAuth func(srcAddr, dstAddr net.Addr, protocol, username, realm string, method string, verdict bool)
  OnAllocationCreated func(srcAddr, dstAddr net.Addr, protocol, username, realm string, requestedPort int)
  OnAllocationDeleted func(srcAddr, dstAddr net.Addr, protocol, username, realm string)
  OnAllocationError func(srcAddr, dstAddr net.Addr, protocol, message string)
  OnPermissionCreated func(srcAddr, dstAddr net.Addr, protocol, username, realm string, peer net.IP)
  OnPermissionDeleted func(srcAddr, dstAddr net.Addr, protocol, username, realm string, peer net.IP)
  OnChannelCreated func(srcAddr, dstAddr net.Addr, protocol, username, realm string, peer net.Addr, channelNumber uint16)
  OnChannelDeleted func(srcAddr, dstAddr net.Addr, protocol, username, realm string, peer net.Addr, channelNumber uint16)
}
```

It is OK to handle only a subset of the events.

Caveats:
- Event handlers are called synchronously, so long-running callbacks will block the server.
- Theoretically, allocations are allowed to change credentials during their lifetime, in that any `Refresh`, `CreatePermission` or `ChannelBind` message may use a different credential. Since this does not seem to be easy to implement with the `PeerConnection` API, credential swaps are currently no tracked.

### DIFF
--- a/internal/allocation/allocation.go
+++ b/internal/allocation/allocation.go
@@ -35,6 +35,8 @@ type Allocation struct {
 	channelBindings     []*ChannelBind
 	lifetimeTimer       *time.Timer
 	closed              chan any
+	username, realm     string
+	eventHandler        EventHandler
 	log                 logging.LeveledLogger
 
 	// Some clients (Firefox or others using resiprocate's nICE lib) may retry allocation
@@ -45,13 +47,19 @@ type Allocation struct {
 }
 
 // NewAllocation creates a new instance of NewAllocation.
-func NewAllocation(turnSocket net.PacketConn, fiveTuple *FiveTuple, log logging.LeveledLogger) *Allocation {
+func NewAllocation(
+	turnSocket net.PacketConn,
+	fiveTuple *FiveTuple,
+	eventHandler EventHandler,
+	log logging.LeveledLogger,
+) *Allocation {
 	return &Allocation{
-		TurnSocket:  turnSocket,
-		fiveTuple:   fiveTuple,
-		permissions: make(map[string]*Permission, 64),
-		closed:      make(chan any),
-		log:         log,
+		TurnSocket:   turnSocket,
+		fiveTuple:    fiveTuple,
+		permissions:  make(map[string]*Permission, 64),
+		closed:       make(chan any),
+		eventHandler: eventHandler,
+		log:          log,
 	}
 }
 
@@ -82,6 +90,14 @@ func (a *Allocation) AddPermission(perms *Permission) {
 	a.permissions[fingerprint] = perms
 	a.permissionsLock.Unlock()
 
+	if a.eventHandler.OnPermissionCreated != nil {
+		if u, ok := perms.Addr.(*net.UDPAddr); ok {
+			a.eventHandler.OnPermissionCreated(a.fiveTuple.SrcAddr, a.fiveTuple.DstAddr,
+				a.fiveTuple.Protocol.String(), a.username, a.realm,
+				a.RelayAddr, u.IP)
+		}
+	}
+
 	perms.start(permissionTimeout)
 }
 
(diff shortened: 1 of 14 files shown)


rg0now:
CC @renandincer

renandincer:
I just looked at it briefly @rg0now! Thank you for PRing this in. 

I think it might be useful to get a handle to the PacketConn from OnAllocatedCreated as well. This might be useful for rate limiting. What do you think?

rg0now:
I think a better way to handle rate limiting would be to provide a [`RelayAddressGenerator.AllocatePacketConn`](https://pkg.go.dev/github.com/pion/turn/v4#RelayAddressGenerator) to the server that will emit rate-limited packetconns. This works even today. The only issue with that approach is if we need user context for applying rate limiting (i.e., when different users may get different rate limits), since currently we miss the user context in the `RelayAddressGenerator` interface handlers. But then again, it would somehow be more adequate to add the user context there. Wdyt?

rg0now:
While working with the lifecycle event handler API it came to our attention that we never report the allocation's relay address in the callbacks. This makes it difficult to track server port allocation status, open firewalls, etc.

The last patch updates the relevant callbacks to also pass the relay address to the caller. The changed event handler signatures are as follows:

```
type EventHandlers struct {
        ...
	OnAllocationCreated func(srcAddr, dstAddr net.Addr, protocol, username, realm string, relayAddr net.Addr, requestedPort int)
        ...
	OnChannelCreated func(srcAddr, dstAddr net.Addr, protocol, username, realm string, relayAddr, peer net.Addr, channelNumber uint16)
	OnChannelDeleted func(srcAddr, dstAddr net.Addr, protocol, username, realm string, relayAddr, peer net.Addr, channelNumber uint16)
}
```

rg0now:
Further evolving the PR based on operational experience: It seems the `PermissionCreated`/`PermissionDeleted` callbacks could also use the relay-address so we added these to the API. The changes:

```
type EventHandlers struct {
        ...
	OnPermissionCreated func(srcAddr, dstAddr net.Addr, protocol, username, realm string, relayAddr net.Addr, peer net.IP)
	OnPermissionDeleted func(srcAddr, dstAddr net.Addr, protocol, username, realm string, relayAddr net.Addr, peer net.IP)
        ...
}
```

rg0now:
We have been deploying this PR into production for a while now, so far all goes as expected. If there are no objections, I'd like to merge this ASAP. Please speak up now if you have concerns.

rg0now:
@JoeTurki Any chance for this to get merged? The code has been merged into [STUNner](https://github.com/l7mp/stunner) for quite some time now and it has been tested by a sizeable user base, so far without any issues reported. Also there was not much opposition here, except the event dispatcher design and some minor stylistic issues (see comments in #423) that I'm happy to fix.

JoeTurki (approved):
I looked at this PR few times actually, it looks great, My only comment is that some callbacks could execute under write-lock, I don't know if this is fixable.
The other comment is that we probably need a callback library under pion or pattern, I talked about this before, but I think this can be merged for now :) 

Thank you :)

rg0now:
@JoeTurki I've rebased the commit, removed the dispatch logic (it was a stupid idea to begin with), and fixed all failing tests. pls merge if you think this is OK.

PR #422, opened 2024-11-18
User-aware relay address generator

### LINKED ISSUE #420: Allow per-user TURN relay allocation
## Summary
Current TURN server API allows us to customize how to allocate and create a net.PacketConn via ManagerConfig.AllocatePacketConn. This callback method however does not provide information of the TURN users for which the allocation was requested.

## Motivation
TURN server is a middlebox where we could enforce policies such as how WebRTC traffic should be relayed. It could also be great place to run various diagnostic services. Allowing these services to be customizable per-user basis creates many business opportunities.

Per-process customization is possible but it is clunky and less flexible.

## Additional context

```mermaid
classDiagram
    TURNServer --> UserManager: AuthHandler(username, realm)  
    TURNClient --> TURNServer: Allocate()  
    TURNServer --> RelayAddressGenerator: AllocatePacketConn(username)
    RelayAddressGenerator --> UserManager: LookupUser(username)

    namespace user_code {
        class UserManager {
        }
        class RelayAddressGenerator {
        }
    }
end
```

enobufs:
Resolves #420

#### Description
Details of this feature are described in #420.
This PR make following changes:
* Added `username` argument to AllocatePacketConn and AllocateConn callback methods in RelayAddressGenerator
* Added/updated corresponding unit tests.

NOTCE: _This is a breaking change for TURN server applications._

#### Reference issue
Fixes #420


server_config.go
```diff
@@ -20,10 +20,10 @@ type RelayAddressGenerator interface {
 	Validate() error
 
 	// Allocate a PacketConn (UDP) RelayAddress
-	AllocatePacketConn(network string, requestedPort int) (net.PacketConn, net.Addr, error)
+	AllocatePacketConn(network string, requestedPort int, username string) (net.PacketConn, net.Addr, error)
```
rg0now, line 23: I'm afraid we will need to add the realm too. From RFC8856:
> Realm:
>    A string used to describe the server or a context within the server. The realm tells the client which username and password combination to use to authenticate requests.

enobufs: Thanks @rg0now for the comment. I thought about it too but because a pion/turn's turn.Server instance can only have one realm, once the allocate request is authorized by the AuthHandler (which would consume the realm from the client), the relay address allocation would not need it because at that point, USERNAME is sufficient to identify the requesting (and authorized) user IMO. I maybe missing something. Please let me know if you had any more thoughts.

rg0now: Fair enough. But then again, due to the same reason (one realm per server) we would not need the `realm` in the `authHandler` either. I think the idea is that for implementations running multiple server instances we can still have a single `authHandler` function instead of having to maintain a separate one for every server instance. But I agree that the case here is somewhat subtle.

internal/allocation/allocation.go
```diff
@@ -28,6 +28,7 @@ type Allocation struct {
 	Protocol            Protocol
 	TurnSocket          net.PacketConn
 	RelaySocket         net.PacketConn
+	Username            string
```
rg0now, line 31: Do we need to export the user name here?

internal/server/turn.go
```diff
@@ -25,7 +25,7 @@ func handleAllocateRequest(r Request, m *stun.Message) error {
 	//    mechanism of [https://tools.ietf.org/html/rfc5389#section-10.2.2]
 	//    unless the client and server agree to use another mechanism through
 	//    some procedure outside the scope of this document.
-	messageIntegrity, hasAuth, err := authenticateRequest(r, m, stun.MethodAllocate)
+	messageIntegrity, username, hasAuth, err := authenticateRequest(r, m, stun.MethodAllocate)
```
rg0now, line 28: Minor stylistic comment: we call `authenticateRequest` from 4 message handlers (allocation request, permission request, channel binding, refresh), but only the first handler needs the username. So I believe it is less intrusive to simply re-parse the username/realm in the allocation handler instead of dropping what's returned at the other 3 invocation sites (see an example in the lifecycle API PR [here](https://github.com/pion/turn/pull/419/files#diff-07ad04f9b305b69b46da66df5d99e7f712864ba650111416305a3fe577f35db1)).

server_config.go
enobufs: > we would not need the realm in the authHandler either.

It is up to the implementor of the authHandler but I would check if the realm matches, not only because TURN client has to reply with the same realm in the challenge (with 401) as required, but also the client might be misbehaving and sending the request to a wrong server.

internal/allocation/allocation.go
enobufs: Good point.  Let me make it private.

internal/server/turn.go
enobufs: Are you sure you'd want to re-iterate over attribute list (linear scan) for the reparse you are suggesting just for the stylistic goal? STUN is small (and sent infrequently) so I don't feel it strongly, but I think it make sense to reuse the artifacts the authorization has just been made with.

rg0now: I'm fine either way, but if we go with your original approach then can you please make it so that we return both the username and the realm from `authenticateRequest` ? The [lifecycle callbacks](https://github.com/pion/turn/pull/419/) that are to be called later during processing the allocate request will need both and it would be an overkill to reparse it (currently we reparse both).

enobufs: In the rebased revision (5cec847):
* Defined authenticationResult to bundle message integrity, username, realm and hasAuth returned by `authenticateRequest`
* Added unit tests for `authenticateRequest`

rg0now: Perfect!

rg0now (approved):

rg0now:
Thanks a lot, this makes the code much leaner. I guess we'll have to wait for @Sean-Der to merge this though because API changes will require a new release (I guess we'll be at v5 already!).

jech:
The basic idea is sound, but we're breaking backwards compatibility for a fairly ad hoc feature.  If we're going to break compatibility, why don't we do it right, so that we don't need to break it again in the future?

I would suggest defining a new type:
```
type TurnContext struct {
    Username string
}
```
and passing that as the new argument.  This way, if we ever feel the need to add extra parameters, we can just add them to `TurnContext` and not break compatibility again.

enobufs:
Agreed. I will update the code.

enobufs:
As I was making changes, i had a couple of thoughts:

### Name the context as `AllocationContext` instead of TurnContext
I feel it is more descriptive for what it does. @jech 

### Add Real field to the context as well
@rg0now I remember you suggested this earlier and I now feel it is good to have it as the context.

I will push my local changes shortly. Thanks.

enobufs:
"allocation context" feel right to me, but I agree that it could be confused by context.Context.

I had following candidate in my head:
* allocation.AllocationContext (stuttering but might not be too bad)
* allocation.UserInfo
* allocation.Metadata
* allocation.Session

(I will fix the lint error later - I have just noticed)

enobufs:
Updated the code, and fixed the lint error also.
@jech I chose "Metadata". It is concise and generic enough to accommodate other possible information we may have in the future.

jech:
Metadata is fine.  (Although frankly, why not just use `data`?  What's so meta about it?)

Perhaps you could squash the commits, it would make review easier?

enobufs:
Thanks @jech, I have squashed the commits.

I feel "metadata" is a suitable term because it represents contextual information related to the operation. It is not primary data, such as _network_ or _port number_, nor is it a parameter required for the operation. Instead, it serves as supplemental information.

joeturki:
Sorry for causing conflicts in your PRs due to the linter updates. Would you like me to fix them for your branch? :)


https://github.com/pion/.goassets/pull/220

Sean-Der:
Now that we have Allocation Event Handlers I believe this use case is covered!

In the example [simple-quota](https://github.com/pion/turn/blob/master/examples/turn-server/simple-quota/main.go#L85) we are able to see usernames coming in and accept/reject as we wish.

Keep me honest @rg0now @adrianosela :) mind double checking?

PR #423, opened 2024-11-18
Add a quota handler callback

rg0now:
Note: the PR is on top of https://github.com/pion/turn/pull/419, see the changes on top of that PR [here](https://github.com/l7mp/turn/pull/3/files).

RFC 8656 includes the following text:

> To mitigate either intentional or unintentional denial-of-service attacks against the server by clients with valid usernames and passwords, it is RECOMMENDED that the server impose limits on both the number of allocations active at one time for a given username and on the amount of bandwidth those allocations can use. The server should reject new allocations that would exceed the limit on the allowed number of allocations active at one time with a 486 (Allocation Quota Exceeded [sic!]) [error].

(Note the mistake in the text: the name of the error is in fact "Allocation Quota Reached", not "Allocation Quota Exceeded".)

This PR adds a quota handler callback function which, if specified, is called by the server just before making an allocation for a user. The handler should return a single bool: if true then the allocation request can proceed, otherwise the request is rejected with the 486 (Allocation Quota Reached) error. Then, the [lifecycle API](https://github.com/pion/turn/pull/419/) can be used to track the number of active allocations per user and this callback can be leveraged to reject allocation requests that would exceed the user's quota.

Note that the other DoS mitigation recommendation given in the RFC (limiting the amount of bandwidth a single user can use) is not targeted by this PR.

### DIFF
--- a/internal/allocation/allocation.go
+++ b/internal/allocation/allocation.go
@@ -35,6 +35,8 @@ type Allocation struct {
 	channelBindings     []*ChannelBind
 	lifetimeTimer       *time.Timer
 	closed              chan any
+	username, realm     string
+	eventHandler        EventHandler
 	log                 logging.LeveledLogger
 
 	// Some clients (Firefox or others using resiprocate's nICE lib) may retry allocation
@@ -45,13 +47,19 @@ type Allocation struct {
 }
 
 // NewAllocation creates a new instance of NewAllocation.
-func NewAllocation(turnSocket net.PacketConn, fiveTuple *FiveTuple, log logging.LeveledLogger) *Allocation {
+func NewAllocation(
+	turnSocket net.PacketConn,
+	fiveTuple *FiveTuple,
+	eventHandler EventHandler,
+	log logging.LeveledLogger,
+) *Allocation {
 	return &Allocation{
-		TurnSocket:  turnSocket,
-		fiveTuple:   fiveTuple,
-		permissions: make(map[string]*Permission, 64),
-		closed:      make(chan any),
-		log:         log,
+		TurnSocket:   turnSocket,
+		fiveTuple:    fiveTuple,
+		permissions:  make(map[string]*Permission, 64),
+		closed:       make(chan any),
+		eventHandler: eventHandler,
+		log:          log,
 	}
 }
 
@@ -82,6 +90,14 @@ func (a *Allocation) AddPermission(perms *Permission) {
 	a.permissions[fingerprint] = perms
 	a.permissionsLock.Unlock()
 
+	if a.eventHandler.OnPermissionCreated != nil {
+		if u, ok := perms.Addr.(*net.UDPAddr); ok {
+			a.eventHandler.OnPermissionCreated(a.fiveTuple.SrcAddr, a.fiveTuple.DstAddr,
+				a.fiveTuple.Protocol.String(), a.username, a.realm,
+				a.RelayAddr, u.IP)
+		}
+	}
+
 	perms.start(permissionTimeout)
 }
 
@@ -90,6 +106,26 @@ func (a *Allocation) RemovePermission(addr net.Addr) {
 	a.permissionsLock.Lock()
 	defer a.permissionsLock.Unlock()
 	delete(a.permissions, ipnet.FingerprintAddr(addr))
+
+	if a.eventHandler.OnPermissionDeleted != nil {
+		if u, ok := addr.(*net.UDPAddr); ok {
+			a.eventHandler.OnPermissionDeleted(a.fiveTuple.SrcAddr, a.fiveTuple.DstAddr,
+				a.fiveTuple.Protocol.String(), a.username, a.realm,
+				a.RelayAddr, u.IP)
+		}
+	}
+}
+
+// ListPermissions returns the permissions associated with an allocation.
+func (a *Allocation) ListPermissions() []*Permission {
+	ps := []*Permission{}
+	a.permissionsLock.RLock()
+	defer a.permissionsLock.RUnlock()
+	for _, p := range a.permissions {
+		ps = append(ps, p)
+	}
+
+	return ps
 }
 
 // AddChannelBind adds a new ChannelBind to the allocation, it also updates the
@@ -114,6 +150,12 @@ func (a *Allocation) AddChannelBind(chanBind *ChannelBind, lifetime time.Duratio
 
 		// Channel binds also refresh permissions.
 		a.AddPermission(NewPermission(chanBind.Peer, a.log))
+
+		if a.eventHandler.OnChannelCreated != nil {
+			a.eventHandler.OnChannelCreated(a.fiveTuple.SrcAddr, a.fiveTuple.DstAddr,
+				a.fiveTuple.Protocol.String(), a.username, a.realm,
+				a.RelayAddr, chanBind.Peer, uint16(chanBind.Number))
+		}
 	} else {
 		channelByNumber.refresh(lifetime)
 
@@ -131,6 +173,12 @@ func (a *Allocation) RemoveChannelBind(number proto.ChannelNumber) bool {
 
 	for i := len(a.channelBindings) - 1; i >= 0; i-- {
 		if a.channelBindings[i].Number == number {
+			if a.eventHandler.OnChannelDeleted != nil {
+				a.eventHandler.OnChannelDeleted(a.fiveTuple.SrcAddr, a.fiveTuple.DstAddr,
+					a.fiveTuple.Protocol.String(), a.username, a.realm,
+					a.RelayAddr, a.channelBindings[i].Peer, uint16(a.channelBindings[i].Number))
+			}
+
 			a.channelBindings = append(a.channelBindings[:i], a.channelBindings[i+1:]...)
 
 			return true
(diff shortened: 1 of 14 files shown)


joeturki:
Sorry for causing conflicts in your PRs due to the linter updates [1]. Would you like me to fix them for all of your branches?

-----

On a side note, I’ve been reviewing past contributions, to understand more about Pion, and I’m a big fan of your reviews. I'm going to get more involved with Pion and would love to submit PRs to implement some of the missing TURN specs (after I finish my todolist for other repos). I’m really looking forward to getting your reviews, can't wait :)

Thank you for all the work you do on Pion/Turn

1. https://github.com/pion/.goassets/pull/220

rg0now:
@joeturki Just fixed the conflicts, no worries. I super-appreciate your work on pion, what you're doing is so immensely useful! If we are at it, can you please also take a look at https://github.com/pion/turn/pull/419 ? 

And thanks for the nice words too. Whenever I have time I come here and do core-reviews. My only worry is that sometimes we may get too picky and this can scare people off (I guess this is how we lost @AndyLc , see https://github.com/pion/turn/pull/315). I guess sometimes we should just let unfinished code in and fix stuff later in order to encourage people to contribute...

joeturki:
Thanks :)

> can you please also take a look at https://github.com/pion/turn/pull/419 ?

I did actually take a look, I do like it. 

My main perspective on this is from a developer experience (DX) standpoint as a library user.
And going through all the Pion repos, I believe we need an abstraction for defining events, systematically:

1. Blocking and behavior-changing events (like the control mechanisms in your PR), That allow us to change a setting or cancel an action etc.
2. Non-blocking telemetry events that allow users to inspect ongoing activity.
3. Events can send custom state.

Something like an event emitter library could help establish consistent behavior for library users, similar to the web EventEmitter, but more tailored to Pion's, I think we need this.

However, I'm not sure, I can starting working on drafting this right now and we can consider postponing it until v5, but I believe we can introduce this without making breaking changes.

> My only worry is that sometimes we may get too picky and this can scare people off (I guess this is how we lost @AndyLc , see https://github.com/pion/turn/pull/315).

Feel free to go hard on me as you can, I just want to do things right :)

cc @jech maybe this is something we should add to the v5 wiki?

server_config.go
```diff
@@ -108,6 +109,87 @@ func GenerateAuthKey(username, realm, password string) []byte {
 	return h.Sum(nil)
 }
 
+// EventHandlers is a set of callbacks that the server will call at certain hook points during an
+// allocation's lifecycle. All events are reported with the context that identifies the allocation
+// triggering the event (source and destination address, protocol, username and realm used for
+// authenticating the allocation). It is OK to handle only a subset of the callbacks.
+type EventHandlers struct {
```
jech, line 116: Should this be called `Callbacks` rather than `EventHandlers`?

Also, should we call it `ServerCallbacks`?  It's not going to appear in a lot of code, perhaps we should be more explicit?

jech:
As a general rule, I'm suspicious of callbacks, I prefer direct style whenever possible, especially in Go.  In this commit, you're adding a whole bunch of callbacks, so naturally I'm suspicious a whole bunch ;-)

Are all of these callbacks necessary?  Perhaps some of these can be omitted, and others can be converted to direct code?

internal/allocation/allocation.go
```diff
@@ -82,6 +90,21 @@ func (a *Allocation) AddPermission(perms *Permission) {
 	a.permissions[fingerprint] = perms
 	a.permissionsLock.Unlock()
 
+	if a.callback != nil {
+		if u, ok := perms.Addr.(*net.UDPAddr); ok {
+			a.callback(EventHandlerArgs{
```
jech, line 95: This code repeats a bunch of times.  Factor it out into a helper function?

internal/server/server.go
```diff
@@ -27,7 +27,8 @@ type Request struct {
 	NonceHash         *NonceHash
 
 	// User Configuration
-	AuthHandler func(username string, realm string, srcAddr net.Addr) (key []byte, ok bool)
+	AuthHandler  func(username string, realm string, srcAddr net.Addr) (key []byte, ok bool)
+	QuotaHandler func(username string, realm string, srcAddr net.Addr) (ok bool)
```
jech, line 31: I think you should be returning `error` here.

server_config.go
```diff
@@ -108,6 +109,87 @@ func GenerateAuthKey(username, realm, password string) []byte {
...
+	OnChannelCreated func(srcAddr, dstAddr net.Addr, protocol, username, realm string,
+		relayAddr, peer net.Addr, channelNumber uint16)
+	// OnChannelDeleted is called after a channel has been removed from the server. The relay
+	// address, the peer address and the channel number can be used to uniquely identify the
+	// channel deleted.
+	OnChannelDeleted func(srcAddr, dstAddr net.Addr, protocol, username, realm string,
+		relayAddr, peer net.Addr, channelNumber uint16)
+}
+
+func genericEventHandler(handlers EventHandlers) allocation.EventHandler { //nolint:cyclop
```
jech, line 150: Why this dispatch function?  It looks to me like complexity for the sake of complexity.  Why not have a bunch of individual handlers stored in a record, and directly call them, without need for the `arg.Type` field?

joeturki:
> As a general rule, I'm suspicious of callbacks, I prefer direct style whenever possible, especially in Go. In this commit, you're adding a whole bunch of callbacks, so naturally I'm suspicious a whole bunch ;-)

I do agree, But what design pattern do you suggest for making overridable / controllable behavior? Should we expose more of the library and make users define their own `Allocation` here for example?

internal/server/server.go
rg0now: The reason why we decided for this API is (1) the error returned by the user is largely ignored by the server (we always [return](https://github.com/l7mp/turn/blob/28967fafea780757536c2a41463994d01924db2a/internal/server/turn.go#L160) the same standard STUN error code `stun.CodeAllocQuotaReached` to the client regardless of the error) and (2) because this is very similar to the `AuthHandler` API. Happy to change though.

jech:
First of all, we shold consider which of these callbacks are actually needed.  My suspicion is that there's only one or two that are actually useful, and that the others are just here for completeness.  We should only implement the ones that are provably needed, since we can always add more later, but once we've added them, we won't be able to remove them.

Second, we should consider whether the full generality of a callback is required.  Perhaps we can just add a boolean or integer option somewhere, and avoid a callback?

server_config.go
rg0now: Since this is top-level API I guess we'll always use this as `turn.XXX`  from the caller site. This suggests that maybe `ServerCallbacks` would be the right name, WDYT?

server_config.go
rg0now: The `EventHandlers` struct here is a top-level API that we must access across several subpackages without creating circular imports. I see 3 ways to solve this:
- Define the struct in a subpackage and then re-export from `server_config.go`. This does not feel idiomatic in pion, plus we lose the top-level docs (Kubernetes does this a lot though, see https://github.com/kubernetes-sigs/controller-runtime/blob/v0.20.0/alias.go).
- Repeat the same def in a stub subpackage that we import everywhere we need it, as well as in the top level `server_config.go`. This just feels wrong though.
- The "dispatcher" approach above. I agree this is ugly but at least it avoids the above isues.

Happy to adopt approach you see fit.

rg0now:
> > As a general rule, I'm suspicious of callbacks, I prefer direct style whenever possible, especially in Go. In this commit, you're adding a whole bunch of callbacks, so naturally I'm suspicious a whole bunch ;-)
> 
> I do agree, But what design pattern do you suggest for making overridable / controllable behavior? 

I think your distinction between callbacks that implement "overridable/controllable behavior" and ones used for telemetry is crucial. For the former (like the auth-handler or the quota-handler) synchronous callbacks seem the best fit, while for the latter (like the event-handler API in https://github.com/pion/turn/pull/419) something asynchronous (like a channel for emitting events) feels way more "right" than the synchronous callbacks we propose above.

Oh, and I agree: pion needs a generic event emitter lib for generalizing this. Do you have any good reference design from a major Go lib we could copy? We could try to prototype the API here in a branch and then, once we're happy with the design we can move it into a separate module.

rg0now:
> First of all, we shold consider which of these callbacks are actually needed. My suspicion is that there's only one or two that are actually useful, and that the others are just here for completeness. We should only implement the ones that are provably needed, since we can always add more later, but once we've added them, we won't be able to remove them.
> Second, we should consider whether the full generality of a callback is required. Perhaps we can just add a boolean or integer option somewhere, and avoid a callback?

Essentially all the server events defined in https://github.com/pion/turn/pull/419 have been asked or proposed in some way or another earlier: authentication events in https://github.com/pion/turn/pull/402, allocation lifecycle events in https://github.com/pion/turn/issues/324, channel lifecycle events in https://github.com/pion/turn/pull/360, etc., and our OpenTelemetry reference design uses each. 

In general I see no reason *not* to be liberal in defining our events: the hooks come with zero runtime overhead and the more data we surface with events the more future proof pion becomes.

jech:
> Essentially all the server events defined in #419 have been asked or proposed in some way or another earlier:

Oh, good.

> In general I see no reason not to be liberal in defining our events: the hooks come with zero runtime overhead and the more data we surface with events the more future proof pion becomes.

If we define a callback now, and then it turns out we need more data in the callback, we cannot easily add a new parameter without breaking compatibility.

If we refrain from defining a callback until there's a demonstrated need for it, there's an increased chance that we'll get it right.

internal/server/server.go
rg0now: > I think you should be returning error here. There should be a specific error value (tested with error.Is) that causes the server to return Quota Exceeded, and the other values should cause it to return Internal server error (or the TURN equivalent, I haven't checked).

`AuthHandler` has the exact same structure in that it returns a bool instead of an error. There's an [in-progress PR](https://github.com/pion/turn/pull/415/files) to redo the design, should we change the signature there to likewise return an error?

JoeTurki (approved):

rg0now:
@JoeTurki I've rebased on top of the latest lifecycle-API PR and fixed all failing tests. pls merge if you think this is OK.

PR #430, opened 2024-12-31
fix: Remove trailing newlines from log messages

levaitamas:
#### Description
This PR fixes a recent linter error that breaks recent PR checks ([an example](https://github.com/pion/turn/actions/runs/12559678074/job/35015821842?pr=429#step:7:5))


levaitamas:
Hi @Sean-Der, can you please take a look at this PR?

joeturki (approved):
Thanks, good catch :)

Issue #438, opened 2025-02-13
Data Race in internal/client/udp_conn.go


jaystenmark:
### Your environment.
  - Version: 31282d5f6ea810f5d98de8e6451c05ff53232355
  - Browser: NA

### What did you do?
I cannot disclose the original test that led to discovering this data race, but I can take the existing test at `internal/client/udp_conn_test.go` and insert a 301 second sleep (anything greater than the 5 minute refresh interval) after `binding.setState(bindingStateReady)` on line 65 to induce the data race, i.e.

```
...
    binding := bm.create(addr)
    binding.setState(bindingStateReady)
    time.Sleep(301*time.Second) // Add this line and package "time" to your imports

    conn := UDPConn{
...
```

Now run `go test -race` in the internal/client directory, wait 5 minutes, and see the data race.

### What did you expect?
No data race should happen.

### What happened?
```
$ go test -race                                                                                                                                 22:36:01 [136/1845]
==================                                                                                                                                                                                                  
WARNING: DATA RACE                                                                                                                                                                                                  
Write at 0x00c00020e860 by goroutine 33:
  github.com/pion/turn/v4/internal/client.TestTCPConn.func4.1()
      /usr/local/google/home/stenmark/pion_mod/turn/internal/client/tcp_conn_test.go:177 +0x1ed
  github.com/pion/turn/v4/internal/client.(*mockClient).PerformTransaction()
      /usr/local/google/home/stenmark/pion_mod/turn/internal/client/client_test.go:28 +0xcf
  github.com/pion/turn/v4/internal/client.(*allocation).CreatePermissions()
      /usr/local/google/home/stenmark/pion_mod/turn/internal/client/udp_conn.go:370 +0x7c1
  github.com/pion/turn/v4/internal/client.(*allocation).refreshPermissions()
      /usr/local/google/home/stenmark/pion_mod/turn/internal/client/allocation.go:126 +0x64
  github.com/pion/turn/v4/internal/client.(*allocation).onRefreshTimers()
      /usr/local/google/home/stenmark/pion_mod/turn/internal/client/allocation.go:159 +0x2c4
  github.com/pion/turn/v4/internal/client.(*allocation).onRefreshTimers-fm()
      <autogenerated>:1 +0x3d
  github.com/pion/turn/v4/internal/client.(*PeriodicTimer).Start.func1()
      /usr/local/google/home/stenmark/pion_mod/turn/internal/client/periodic_timer.go:52 +0x130

[...]

PR #439, opened 2025-02-13
Fix data race in WriteTo

jaystenmark:
#### Description
There is no harm in shadowing `err` in the goroutine since we do not need it outside of the routine.

I also discovered data races in tcp_conn_test.go while fixing this bug and have included them in this PR.

#### Reference issue
Fixes [#438 ](https://github.com/pion/turn/issues/438)


Sean-Der:
I am going to re-open these PRs. CI was broken, and will be working on new PR.

sorry about that!

PR #447, opened 2025-03-26
Update lint rules, use testify/assert

RyanPotat:
#### Description

Update lint rules to forbid the use of direct testing.T helpers. Removed a few `//nolint:cyclop` only by consequence of using assert. 🦆 

https://github.com/pion/.goassets/pull/222

Reference issue

N/A


JoeTurki (approved):
Thank you so much Ryan

PR #450, opened 2025-04-15
Handle nil TURN addresses in retransmissions

amanakin:
#### Description
Handle nil TURN addresses in retransmissions

In certain scenarios, the `To` field of a transaction can be nil (nil TURN address of the `Client`).

### DIFF
--- a/client.go
+++ b/client.go
@@ -662,7 +662,7 @@ func (c *Client) onRtxTimeout(trKey string, nRtx int) {
 	}
 
 	c.log.Tracef("Retransmitting transaction %s to %s (nRtx=%d)",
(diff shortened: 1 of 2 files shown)


Sean-Der:
This LGTM @amanakin!

Would it be possible to add a unit test to this? This change could easily be undone, and start causing you to break again :/

amanakin:
Good point @Sean-Der! Added tests for happy-path with and w/o retransmissions. 
Couldn't find a better way to cause timeout, only setting 1ns rto.

Sean-Der (approved):

Sean-Der:
Thank you @amanakin :)

PR #455, opened 2025-05-07
Add client config param IgnoreTURNResolveErrors

amanakin:
#### Description
In some configurations we don't use the TURN address at the client side (e.g. proxy connection, see https://github.com/pion/turn/pull/450). But now at the client setup the provided turn address must be resolvable. 

Adding client config param `IgnoreTURNResolveErrors` that allows omit resolving errors when we don't need turn address. 

Also discussed here: https://github.com/pion/ice/pull/773

### DIFF
--- a/client.go
+++ b/client.go
@@ -47,6 +47,8 @@ type ClientConfig struct {
 	Conn           net.PacketConn // Listening socket (net.PacketConn)
 	Net            transport.Net
 	LoggerFactory  logging.LoggerFactory
+
+	IgnoreTURNResolveErrors bool // TURN server address is not required for some configurations (e.g. proxy)
 }
 
 // Client is a STUN server client.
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -47,6 +47,8 @@ type ClientConfig struct {
 	Conn           net.PacketConn // Listening socket (net.PacketConn)
 	Net            transport.Net
 	LoggerFactory  logging.LoggerFactory
+
+	IgnoreTURNResolveErrors bool // TURN server address is not required for some configurations (e.g. proxy)
```
JoeTurki, line 51: I don't think we should add a config, there are many reason, but it feels like a single use flag.
What if we make a new constructor `NewClientWithResolver` and essentially make the nested condition after `	if len(config.TURNServerAddr) > 0 {` customizable, we can then make the current `NewClient` call this new constructor with the old resolver logic.

JoeTurki: I can make a PR for this!

amanakin: @JoeTurki So you mean something like 
```
if len(config.TURNServerAddr) > 0 {
	turnServ, err = config.ResolveTURN(config.TURNServerAddr)
	if err != nil {
		return nil, err
	}
}
```
And then in some cases use [there](https://github.com/pion/ice/blob/master/gather.go#L818) `WithResolveTURN(...)` that will return `nil, nil`?

amanakin: @JoeTurki Any updates on this?

PR #456, opened 2025-05-15
Lower priority of per-packet error message

### LINKED ISSUE #197: Large numbers of errors in log
I've just run Galène against pion/turn instead of coturn.  The server was moderately loaded, a few lectures with 20 or so students each.  The log is riddled with messages that indicate problems between Galène and the TURN server — I haven't noticed any that indicate issues with communication with clients.

If my understanding of TURN is correct, some of the errors are not actual errors, but simply normal negotiation between the TURN client and server; in that case, the issue is that the default logging is too verbose.  Others look worrying, and I'd appreciate a clarifiction.

Typical errors include:
```
turnc ERROR: 2021/01/19 16:34:56 fail to refresh permissions: CreatePermission error response (error 441: Wrong credentials: the realm value is incorrect
```
and
```
turn ERROR: 2021/01/19 16:35:01 error when handling datagram: failed to handle CreatePermission-request from 51.210.14.2:37880: no allocation found 51.210.14.2:37880:51.210.14.2:1194
```
and
```
turnc ERROR: 2021/01/19 16:35:09 fail to refresh permissions: all retransmissions failed for yoM0LWftNldeEfRt
```
and
```
turn ERROR: 2021/01/19 16:35:51 error when handling datagram: failed to handle Refresh-request from 51.210.14.2:37880: no allocation found 51.210.14.2:37880:51.210.14.2:1194
```

jech:
The error message "Failed to handle datagram" is generated for
each received packet; it can therefore be used by an attacker to
fill the server's logs by sending a burst of incorrect packets.

Lower its priority to debug.

Fixes #197 and jech/galene#271.

### DIFF
--- a/server.go
+++ b/server.go
@@ -233,7 +233,7 @@ func (s *Server) readLoop(conn net.PacketConn, allocationManager *allocation.Man
 			ChannelBindTimeout: s.channelBindTimeout,
 			NonceHash:          s.nonceHash,
 		}); err != nil {
-			s.log.Errorf("Failed to handle datagram: %v", err)
+			s.log.Debugf("Failed to handle datagram: %v", err)
 		}
 	}
 }


JoeTurki (approved):
Makes sense!

PR #458, opened 2025-06-02
V4.0.0 perm refresh

valley3405:
#### Reference issue
Fixes #...

### DIFF
--- a/internal/client/udp_conn.go
+++ b/internal/client/udp_conn.go
@@ -18,7 +18,7 @@ import (
 
 const (
 	maxReadQueueSize    = 1024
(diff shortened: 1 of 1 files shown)


valley3405:
15s

JoTurk:
Hello, Is there an issue for this? I think 100-150 is the ideal for most uses. But I think we can a add a setting to customize the interval for niche use.

Sean-Der:
Sorry this didn't get address sooner @valley3405!

I have exposed this is as a config option. It still defaults to 120ms, but can be reduced.

PR #460, opened 2025-06-10
RFC: Add a configurable sized nonce generator

rg0now:
The aim of This PR is to reduce the amplification factor in reflection attacks against our TURN servers. To summarize, it seems that recently TURN servers were targeted in large-scale reflection+amplification attacks that abuse the TURN allocation request/response mechanism. See [background](https://medium.com/@ggarciabernardo/reflection-amplification-attacks-abusing-turn-servers-743803828593) and the corresponding [coturn](https://github.com/coturn/coturn/issues/1687) issue.

The default amplification factor, the ratio of the size of the initial Allocation Response message reflected by the server to the victim IP compared to the  size of the initial Allocation Request sent by the malevolent client, is currently about 2.2x for pion/turn. This makes pion TURN servers an appealing target for reflection+amplification attacks. Luckily we do not generate the SOFTWARE attribute by default and by using a 2 byte REALM we can go down to about 2.1x without touching the code. The rest, however, is mostly caused by the 80 byte nonces pion/turn uses.

This PR makes the nonce size configurable and removes some redundancy by using a shorter 4-byte timestamp at one minute precision and a more efficient base36 encoder/decoder. The default is 4 bytes timestamp + 12 byte truncated HMAC hash, which gives about 24 bytes long nonces including the overhead of base36 encoding and padding.

**Warning:** this PR has not been tested against real browsers. The pion/turn TURN client seems to handle the smaller nonces reliably though.

### DIFF
--- a/internal/server/base36.go
+++ b/internal/server/base36.go
@@ -1,0 +1,63 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+package server
+
+import (
+	"math/big"
+	"strings"
+)
+
+// Base36 alphabet for encoding.
+const base36Alphabet = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+
+// EncodeBase36 converts bytes to base36 string using big.Int for arbitrary length.
+func encodeBase36(data []byte) string {
+	if len(data) == 0 {
+		return ""
+	}
+
+	num := new(big.Int).SetBytes(data)
+	if num.Cmp(big.NewInt(0)) == 0 {
+		return "0"
+	}
+
+	base := big.NewInt(36)
+	buf := make([]byte, 0, len(data)*2)
+	remainder := new(big.Int)
+	for num.Cmp(big.NewInt(0)) > 0 {
+		num.DivMod(num, base, remainder)
+		buf = append(buf, base36Alphabet[remainder.Int64()])
+	}
+
+	for i, j := 0, len(buf)-1; i < j; i, j = i+1, j-1 {
+		buf[i], buf[j] = buf[j], buf[i]
+	}
+
+	return string(buf)
+}
+
+// DecodeBase36 converts base36 string back to bytes using big.Int for arbitrary length.
+func decodeBase36(encoded string) []byte {
+	if encoded == "" {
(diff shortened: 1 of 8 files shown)


JoTurk:
This is awesome.

> Warning: this PR has not been tested against real browsers. The pion/turn TURN client seems to handle the smaller nonces reliably though.

Okay, I'm going to test this branch with real browsers.

rg0now:
Redid the PR to pass the linter checks. Always full of excitement. Maybe we should add some tests for the base36 encoder/decoder, I just copy-pasted a quick implementation from an older repo.

rg0now:
I've remade the PR with your modifications, thanks.

rg0now:
Merged this into STUNner dev, no reports so far. Also tested with STUNner+LiveKit+FF, no issues.

JoeTurki (approved):

rg0now:
@JoeTurki I've rebased the commit and added tests for the base36 encoder. pls merge if you think this is OK.

rg0now:
@JoeTurki rebased on top of master, I think this can be safely merged. This would let the new nonce manager to get some broader testing: we have not seen any browser incompatibility yet but you can never know.

rg0now:
We have created a minimal tool to test the amplification factor of different TURN servers, see [here](https://github.com/l7mp/turn-amplification-tool). Below are the results.

| Server                              | Amplification factor |
|-------------------------------------|----------------------|
| undisclosed proprietary TURN server | 4.22x                |
| pion/turn (default)                 | 3.44x                |
| pion/turn (no realm)                | 3.22x                |
| short-nonce (default)               | 1.89x                |
| short-nonce (no realm)              | 1.44x                |
| short-nonce (maclen=2, no realm)    | 1.22x                |

This PR reduces the amplification factor from 3.44x of default pion/turn to 1.44x (using an empty realm which I'm not sure is valid). Security savvy people can enforce `maclen=2` during [server initialization](https://github.com/l7mp/turn/blob/0d702f5ffdeffad5f1440c9f5dfa22acbff04a4e/server.go#L55) to bring down the amplification factor to 1.22x. I think there's room for further improvement (use a fully random nonce, say, 4 bytes), not sure it's worth the pain.

JoTurk:
Perfect, thank you <3

PR #464, opened 2025-07-17
Added user-quota as server config

arindamnayak:
#### Description
user-quota config added for managing turn allocations counts per turn creds

#### Reference issue
This MR introduces user quota functionality to restrict the number of allocations per TURN credentials, similar to coturn's user-quota configuration. The implementation allows end users to set allocation limits per credential, preventing resource exhaustion attacks and unauthorized usage. This enhancement improves security posture by limiting potential abuse while providing feature parity with coturn's quota system. The quota configuration is integrated into the existing credential validation flow, ensuring seamless operation with minimal performance impact.

### DIFF
--- a/internal/allocation/allocation.go
+++ b/internal/allocation/allocation.go
@@ -46,7 +46,12 @@ type Allocation struct {
 }
 
 // NewAllocation creates a new instance of NewAllocation.
-func NewAllocation(turnSocket net.PacketConn, fiveTuple *FiveTuple, log logging.LeveledLogger, username string) *Allocation {
+func NewAllocation(
+	turnSocket net.PacketConn,
+	fiveTuple *FiveTuple,
+	log logging.LeveledLogger,
+	username string,
+) *Allocation {
 	return &Allocation{
 		TurnSocket:  turnSocket,
 		fiveTuple:   fiveTuple,
--- a/internal/allocation/allocation_manager.go
+++ b/internal/allocation/allocation_manager.go
@@ -151,7 +151,7 @@ func (m *Manager) DeleteAllocation(fiveTuple *FiveTuple) {
 	allocation := m.allocations[fingerprint]
 	if m.userCounts != nil && allocation != nil {
 		m.userCounts[allocation.username] -= 1
-		if m.userCounts[allocation.username] == 0 {
+		if m.userCounts[allocation.username] < 1 {
 			delete(m.userCounts, allocation.username)
 		}
 	}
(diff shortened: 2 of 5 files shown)


arindamnayak:
@rg0now , would you able to review this change?

rg0now:
This functionality is also covered by #423, which is on top of the general event handler API proposed in #419. Can you please clarify the relation of this PR with those ones, if any?

In general, I'm not against merging PRs that implement useful functionality in piecemeal deltas instead of the larger PRs (like the event handler as of #419) which are always much more difficult to get merged. That being said, if there's any chance for the event handler API to get merged in a timely fashion then I'd drop this PR and go with the more generic approach in #423.

arindamnayak:
@rg0now , I am also looking for #423 MR to get merge , but this MR has more changes and open since nov 2024. I am not sure if this change can be merged sooner, hence in interim I am looking forward for this change.

rg0now:
Let me try to update #419 and #423 first, hopefully we can get those ones merges soon

arindamnayak:
@rg0now , the only concern is if #419, #423 gets delayed or blocked due to some reason, we should fast track this MR atleast.

rg0now:
I've rebased #419 and #423 on top of the latest master and we got an approval from @JoeTurki so hopefully both will be merged soon.

arindamnayak:
Thanks @rg0now , please follow up to get those MRs merged, accordingly I can update this MR.

JoTurk:
both PRs finally got merged :)

JoTurk:
@rg0now Should we release the next version as v4.1.0 or just v4.0.3? I'd lean toward bumping the minor version and going with v4.1.0.

rg0now:
@JoeTurki I don't have strong opinions on versioning, anything greater than the current will work...:-)

As per this PR: I think it would be extremely useful to re-implement this on top of #419 and #423 (just track the number of allocations per user via the `OnAllocationCreated`  and `OnAllocationDeleted` callbacks) and turn it into an example, wdyt?

JoTurk:
> @JoeTurki I don't have strong opinions on versioning, anything greater than the current will work...:-)

Released as v4.1.0 :)

Sean-Der:
Sorry I am behind on context :(

@rg0now @arindamnayak Is it ok to close this now that we have https://github.com/pion/turn/commit/189f4304 ?

rg0now:
TL;DR Agree, I think we can safely close this. 

That being said, an example demonstrating the use of the current API would be nice, because the quota handler must be integrated with the lifecycle API to reproduce the functionality of this PR: the `OnAllocation*` event handlers should count the number of allocations per username and then the `QuotaHandler` can use that info to accept/reject allocations. The advantage of the `QuotaHandler` approach is its flexibility: you're free to outsource the allocation counters to a DB, implement per-user quotas, or charge users for exceeding their quota. We've implemented this in STUNner and it works like charm. I'll take a note to myself to contribute a simple example.

Sean-Der:
@rg0now I will take a crack at that right now :)

PR #466, opened 2025-08-07
Fix error reporting in SendIndication handler

rg0now:
Fix misleading error reporting when a SendIndication message cannot be relayed. Same pattern also used [here](https://github.com/pion/turn/blob/189f4304be550e7cd0aeab443539f7ba6c13a2ad/internal/server/turn.go#L451).


JoTurk:
Is it possible to add a test case for this, so this behavior is defined?

rg0now:
@JoeTurki Turns out this was much more difficult to test for since peer-side write errors are not reported back to the client. Luckily we now have an allocation lifecycle API so we can use the `OnAllocationError` callback to catch server-side issues...:-)

Wdyt?

PR #481, opened 2025-10-17
Handle stale nonces in ChannelBind, and add timer to renew bindings

### LINKED ISSUE #471: Force use of channels
Hi,

Is there a way to force use of Channels instead of Indications where in practice my app seems to be randomly defaulting either on Indications or Channels at connection start-up.

What are the forces in effect influencing the behavior?

berkant:
#### Description
While the issue was originally about discussing why turn would fall back to indications after a while in a long-lived connection, tracing down into the code revealed stale nonces aren't being handled whatsoever by the `bind()` method, causing channel bindings to time out and transition to failed state, forcing the connection to go 2-way over indications for the rest of its life.

This PR adds code handling stale nonce errors to ChannelBind requests. It also decouples logic related to binding lifecycle (initial binding & binding renewal) from `WriteTo` by carving out code into a timer, similar to the way perms and allocs work.

#### Reference issue
Fixes #471

### DIFF
--- a/internal/client/binding.go
+++ b/internal/client/binding.go
@@ -61,6 +61,12 @@ func (b *binding) refreshedAt() time.Time {
 	return b._refreshedAt
 }
 
+func (b *binding) ok() bool {
+	state := b.state()
+
+	return state == bindingStateReady || state == bindingStateRefresh
+}
+
 // Thread-safe binding map.
 type bindingManager struct {
 	chanMap map[uint16]*binding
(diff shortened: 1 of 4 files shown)


berkant:
@JoeTurki WDYT?

JoTurk:
@berkant Thank you so much, I like this change, I'll review it again in details either tonight or this weekend.

berkant:
@JoeTurki Is this making upstream?

JoTurk:
@rg0now what do you think about this? looks like a good patch to me.

rg0now:
I've started a review like weeks ago, then realized I don't know the next thing about the channel lifecycle, then I realized stale nonces can become an issue, then I got lost on the server side trying to find out how this is implemented there, and then I forgot to come back here.  (BTW, I'm always surprised by how much cleaner the server-side code is *than the client side*.) I think the PR cleans up a lot of things, please merge

PR #500, opened 2025-12-20
Implement section-4.3 from RFC6062

Sean-Der:
Relates to #118


Sean-Der:
@JoeTurki @rg0now I am going to try and implement this with lots of small PRs. Not done yet with this, but wanted to get just [0] working, and then get accepting in a subsequent PR.

We should also allow the user to provide a customer dialer. I saw this suggested in previous PRs. I will mark this with `Co-author` also, lots of people have worked on this :) 

[0] https://datatracker.ietf.org/doc/html/rfc6062#section-4.3

JoTurk:
@Sean-Der This is sooo dope, this was on my todo list for a while, and I feel bad when people ask for features in pion turn and i direct them to coturn or other implementations instead.

rg0now:
I was just wondering how this would change the [`ServerConfig`](https://pkg.go.dev/github.com/pion/turn/v4#ServerConfig). Do you plan a `ListenerConfig` type to mean the server port is a TURN/TCP listener? Or just add a "type" or "protocol" field to all `PacketConnConfigs` and `ListenerConfigs` with sane defaults? Can we encapsulate the peer-side complexity (the same TCP relay address must work both as a TCP server for peers to be able to connect back to the TURN client and a TCP client for the reverse direction to work) into the `RelayAddressGenerator.AllocateConn()`? And all this with a proper `net.Dialer` like interface (this is the point where the last attempt petered out). Just pondering here, no specific opinions

Sean-Der:
@rg0now I am not sure yet!

What do you think of me landing things with zero configuration, and then we add it back when everything is working?

Sean-Der:
@JoeTurki @rg0now Ok can I get a review now?

It's not future complete, but I think this is a good moment to stop and rip into it :) 

* Our TURN Client works against it
* TURN server hardcodes UDP everywhere when dealing with FiveTuple (this was a problem before)
* I need to add 30 second timeouts
* Add things to ServerConfig to give users control
* and probably more!

At least a few more PRs!

JoTurk:
@Sean-Der I'll review and test this :)

internal/allocation/allocation_manager.go
```diff
@@ -244,3 +246,75 @@ func (m *Manager) GrantPermission(sourceAddr net.Addr, peerIP net.IP) error {
...
+	if len(peerAddress.IP) == 0 || peerAddress.Port == 0 {
+		return 0, errInvalidPeerAddress
+	}
+
+	rand64, err := randutil.CryptoUint64()
+	if err != nil {
+		return 0, err
+	}
+
+	conn, err := net.DialTCP("tcp4", nil, &net.TCPAddr{IP: peerAddress.IP, Port: peerAddress.Port}) // nolint: noctx
```
rg0now: This is the critical point: (1) the Dialer should be an abstract interface and (2) the local address:port must be the transport relay address for the allocation (the TCPAddr we create an `allocation.CreateAllocation` once we change that to call `AllocateConn` per the requested transport). The former can be taken care of later, but for the latter we will need to add the `SO_REUSEADDR` (and maybe also the `SO_REUSEPORT`? I always forget the difference) otherwise we cannot bind multiple peer connections to the same local addr:port. AFAIK `SO_REUSEADDR` is supported in all major OSes but `SO_REUSEPORT` is Linux only

rg0now:
> What do you think of me landing things with zero configuration, and then we add it back when everything is working?

+1 for the pragmatic approach

It just occurred to me that the same TCP listener can work both for creating UDP relay connections and TCP relay connections depending on the requested-transport. Consider my request for adding a "protocol" field to `{PacketConn,Listener}Configs` void

Otherwise I think the code is pretty nice already!

Sean-Der:
@rg0now 100% agree with your API. 

What do you think of landing this, and then I will expose an API so users can adjust the Listener + Dialer. Easier to review/make progress, but if that has downsides happy to merge as one big PR!

JoTurk:
Okay i reviewed half of this :) 
will continue tomorrow.

Sean-Der:
Nice, thank you @rg0now and @JoeTurki :D

I will start the next part now. Will open up a dedicated PR that just does configuration and we can take a few passes to make sure every use case works :)

JoTurk:
@Sean-Der I was currently reading the final part and I was going to comment / ask about how we handle 

>   If the server is currently processing a Connect request for this
   allocation with the same XOR-PEER-ADDRESS, it MUST return a 446
   (Connection Already Exists) error.
   
but i guess this is something we're going to do later?

Sean-Der:
Sorry I aggressively merged @JoeTurki!

We protect against dupes here https://github.com/pion/turn/blob/master/internal/server/turn.go#L504-L510 and it should be safe for a single session.

Oh but I don't handle across many sessions! I should put the address in the map earlier, and then delete on failure maybe? I will fix that.

JoTurk:
@Sean-Der no, that's totally fine to merge :) don't wait for me. really excited about more TLS / TCP. If we found an issue in the spec we can fix it later :)

Issue #487, opened 2025-12-08
TURN allocation marshaler/unmarshaler


rg0now:
## Summary

A way to serialize/deserialize TURN allocation state would make it possible to back up TURN allocations in a key-value store. This would let one implement high-availability TURN servers and handle upstream NAT instability.

## Motivation

The prerequisite of implementing high-availability TURN servers is to maintain allocation state in a key-value store like Redis. Then, once a crashing TURN server has restarted (over the same IP) it could recover existing allocations from the key-value store (at least UDP and probably DTLS, not sure about TCP, TLS). This may also let us handle upstream NAT instability (a TURN packet that belongs to an active allocation being re-NAT'd to an unknown 5-tuple), but that would be pretty tricky to implement.

## Describe alternatives you've considered

None. 

## Additional context

See e.g., the [pion/dtls binary marshaler/unmarshaler](https://pkg.go.dev/github.com/pion/dtls/v3#State.MarshalBinary).

rg0now:
Thanks, this is interesting, will look into this soon. Meanwhile, can you please split this large PR into multiple independent PRs so that it's easier to review? E.g.,
- binary marshalers/unmarshalers,
- `Storage` interface with a default in-memory implementation,
- an example in `examples/turn-server` that shows how to use the Storage interface to externalize allocation state into Redis.

rg0now:
It just occurred to me that this comment went to the wrong place. Thanks @Sean-Der for [linking it](https://github.com/pion/turn/pull/504#issuecomment-3693617174).

PR #504, opened 2025-12-25
feat(allocation): Add Serializable and Pluggable Storage for Allocations

Amr-Shams:
[#487] This pull request significantly enhances the TURN server's architecture by introducing two major features: serialization of allocation objects and a pluggable storage backend. 

## Summary

1. Making Core Objects Serializable: Allocation, Permission, ChannelBind, and FiveTuple can now be marshaled to and from a binary format.
2. Introducing a Storage Interface: A Storage interface has been created to abstract the underlying storage mechanism for allocations.
3. Adding a Redis(under testing): a first trial for adding redis for a more robust in memory storage.

## Key Technical Changes

- Serialization: Implemented MarshalBinary and UnmarshalBinary on the core allocation objects, enabling them to be stored and retrieved from external systems.
- Storage Abstraction:
  - A new Storage interface is defined in `internal/allocation/storage.go`.
  - The `AllocationManager` is refactored to use this interface
- Redis Implementation:
  - `internal/allocation/redis_storage.go` concret implementation for storage server.
  - This allows multiple TURN server instances to share a single, consistent state.
- Testing:
  - Added extensive unit tests for the serialization logic to ensure correctness.

## How to Test

- All existing and new unit tests are passing.
- for redis i think we can add more tests

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -8,6 +8,7 @@ package turn
 
 import (
 	"context"
+	"io"
 	"net"
 	"runtime"
 	"sync/atomic"
@@ -421,8 +422,10 @@ func TestTCPClientDial(t *testing.T) {
 	remotePeerConn, err := net.Listen("tcp4", "127.0.0.1:0") // nolint: noctx
 	assert.NoError(t, err)
 
+	hasReadCtx, hasReadCancel := context.WithCancel(context.Background())
+	hasClosedCtx, hasClosedCancel := context.WithCancel(context.Background())
+
 	expectedMsg := []byte{0xDE, 0xAD, 0xBE, 0xEF}
-	ctx, cancel := context.WithCancel(context.Background())
 	go func() {
 		inboundTCPConn, inboundErr := remotePeerConn.Accept()
 		assert.NoError(t, inboundErr)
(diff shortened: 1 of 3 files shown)


Amr-Shams:
though there are some concern changes i have made for the liftimer timer when close is done (it should then expires the expiry time or not? for permssions and when i should do that for the connection

Sean-Der:
This is amazing @Amr-Shams I am so excited you are doing this :)

Check out [this](https://github.com/pion/turn/issues/487#issuecomment-3693201633) comment. 

Would you mind starting with a small PR that just does `MarshalBinary` and `UnmarshalBinary`

Maybe start with something super basic like 'Can a server come back up with the same listening port' we don't even need to worry about allocations yet.

Sean-Der:
I think it might be worse for users to have a Redis dependency/interaction.

Everyone's usage of these systems is so different, so we need to be unopionated.

Amr-Shams:
> I think it might be worse for users to have a Redis dependency/interaction.
> 
> Everyone's usage of these systems is so different, so we need to be unopionated.

thank you for the feedback i have divided the pr to a mini steps according to the comment

PR #506, opened 2025-12-30
Add TCP Relay support

### LINKED ISSUE #118: Implement RFC-6062

Sean-Der:
Resolves #118

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -381,7 +381,7 @@ func TestClientReadTimout(t *testing.T) {
 	assert.Contains(t, err.Error(), "use of closed network connection")
 }
 
-func TestTCPClientDialTCP(t *testing.T) {
+func TestTCPClientDial(t *testing.T) {
 	tcpListener, err := net.Listen("tcp4", "0.0.0.0:3478") //nolint: gosec,noctx
 	require.NoError(t, err)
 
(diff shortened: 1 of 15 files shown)


Sean-Der:
💥 

@JoeTurki @rg0now full support! I can use all the APIs in the client now. I will add more tests tomorrow.

* I need to add 30 second timeout detection
* This requires a new major version, but I think it's the best way do it. I didn't see a way to make `NewConn` work.
* ?

Sean-Der:
@rg0now @JoeTurki Ok mind checking it out! I feel really good about this. 

One or two more PRs of stuff I missed, then it's release worthy.

Issue #427, opened 2024-12-11
Feature request: proxy support


jech:
## Summary

There appears to be no way to configure the TURN client to work through an HTTP or SOCKS5 proxy, nor even over an SSH tunnel.

## Motivation

While working on [galene-stt](https://github.com/jech/galene-stt), I was given access to a server with a powerful GPU.  Unfortunately, the server was firewalled away from the Internet: the only ways it could communicate with the outside world were an HTTP proxy and ssh access.

The obvious solution would have been to configure Pion to use TURN over TCP over HTTP CONNECT.  The other solution would have been to use the built-in SOCKS proxy of ssh.  Unfortunately, there's no way to tell our TURN client to use a proxy.

## Describe alternatives you've considered

I ended up hacking the client code to speak TURN over TCP over an ssh tunnel.  It wasn't pretty.

## Additional context

I think the right way to do that would be to have an implementation of `transport.Net` that uses `golang.org/x/net/proxy`.  Not sure how the client code would request that ­— should it be honouring some environment variable, or should there be an extra option to `pion.NewApi`?

rg0now:
[Copying comment from Slack here for visibility]

My understanding is that the TURN client should work on top of anything that is a `net.Conn`, at least if you are using it directly (I mean, not via pion/ice.) So if you can open the ssh connection or the HTTP proxy and present it to the client as a `net.Conn` then you should be fine. Now of course the server should be able to decapsulate whatever tunnel you're using, so a TURN server listener should exist with  a matching `net.Listener`.

jech:
> My understanding is that the TURN client should work on top of anything that is a `net.Conn`

I could be wrong, but I believe that it requires a `transport.Net`, which is slightly more than a `net.Conn`.

> if you are using it directly (I mean, not via pion/ice.)

The goal is to be able to use a proxy with an application that uses pion/turn through pion/ice through pion/webrtc.

gittethis:
> I ended up hacking the client code to speak TURN over TCP over an ssh tunnel. It wasn't pretty.
> 
hi!, 
would you share the where and how is it possible to hack the client to speak TURN over TCP ? (i'm in a kinda same context and exploring ways to "evade").

Thanks!

rg0now:
[...]

PR #507, opened 2025-12-30
Initialise unknownAttributes to nil

jech:
This avoids allocating a slice header for each packet.

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -87,7 +87,7 @@ func handleTURNPacket(req Request) error {
 		)
 	}
 
-	unknownAttributes := []stun.AttrType{}
+	unknownAttributes := []stun.AttrType(nil)
 	for _, attr := range stunMsg.Attributes {
 		if attr.Type.Required() && !attr.Type.Known() {
 			unknownAttributes = append(unknownAttributes, attr.Type)


Sean-Der (approved):
Thank you @jech !

jech:
@Sean-Der my pleasure.  I'm catching up on six months during which I didn't have time to review the changes to Pion, and a lot has happened!

Sean-Der:
I hope to have some stuff that is even more fun :)

I am going to make it so you can run TURN/HTTP on same port (you put out idea a while back). Will make it easy to use.

Also the ability to serialize server state. So you can restart TURN server (or it can crash) and can come right back up. Excited for you to try it out!

jech:
> I am going to make it so you can run TURN/HTTP on same port (you put out idea a while back).

Cool.  While you're at it, could you please think about #427?  The implementation is easy, the tricky bit is what API we should use so that a WebRTC application can be easily coerced to go through a proxy.

> Also the ability to serialize server state. So you can restart TURN server (or it can crash)

That's what the companies want (they love advertising high-availability even when it's not useful), but I'm more interested in end-to-end solutions.  If your TURN server reboots, then ICE will restart and reestablish the allocations.  All the smarts at the edge, as Louis Pouzin intended.

PR #511, opened 2026-01-03
Add test for client and peer connection address over a TCP alloc

rg0now:
#### Description

For client-initiated connection requests over a TCP allocation the outbound connections made by the server MUST be locally bound to the relay transport address for the allocation. In particular, [RFC 6062 Section 5.2](https://datatracker.ietf.org/doc/html/rfc6062#section-5.2) contains the following text:

> Otherwise, the server MUST initiate an outgoing TCP connection.  The **local endpoint is the relayed transport address associated with the allocation**. 

This PR adds a check for this.

#### Reference issue
See #512

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -427,6 +427,9 @@ func TestTCPClientDial(t *testing.T) {
 		inboundTCPConn, inboundErr := remotePeerConn.Accept()
 		assert.NoError(t, inboundErr)
 
+		remoteAddr := inboundTCPConn.RemoteAddr()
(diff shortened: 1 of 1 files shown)


Sean-Der:
🔥 thanks for this I’m terrible with spec compliance, very sloppy :(

We could just put this under a OS check maybe? For platforms that support REUSE_PORT apply it in dedicated file?

rg0now:
The test is included in #514. Closing this.

PR #514, opened 2026-01-05
Bind relay connections to the relay address for TCP allocations

### LINKED ISSUE #512: Random local address on relay connections over TCP allocations
For client-initiated connection requests over a TCP allocation the outbound connections made by the server MUST be locally bound to the relay transport address for the allocation. In particular, [RFC 6062 Section 5.2](https://datatracker.ietf.org/doc/html/rfc6062#section-5.2) contains the following text:

> Otherwise, the server MUST initiate an outgoing TCP connection. The **local endpoint is the relayed transport address associated with the allocation**.

See #511 for a test for this. That test is currently failing: checking the remote address at the peer we get a random address instead of the relay address (also checked with WireShark). This will break ICE. Note that `clientConn.LocalAddr()` correctly returns the relay address.

The reason is that the [allocation manager `CreateTCPConnection` binds the new connection to a random local address](https://github.com/pion/turn/blob/9e70c937d8dc67b1baec450f2cb3322092f57873/internal/allocation/allocation_manager.go#L288) instead if the relay address.

The fix can be to create the connection over a `net.Dialer` with `SO_REUSEADDR` (allows rebinding after close/TIME_WAIT) and `SO_REUSEPORT` (allows concurrent active binds) enabled. AFAIK both are needed. The problem is that `SO_REUSEPORT` is Unix-only (not sure about current Windows).

Happy to contribute a fix but before that I wanted to hear people's opinion on the OS-support issue.

*Update:* seems [Windows needs only `SO_REUSEADDR`](https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse)

rg0now:
This is a humble attempt to fix #512

The trick is to use the proper sockopts to create the TCP relay listener and the relay connections (`SO_REUSEADDR` and wherever needed, `SO_REUSEPORT`). 

Caveats:
- We import a battle-tested [external module](https://github.com/libp2p/go-reuseport) to hide the OS-dependent "reuseport" sockopts. Let me know if adding new imports is discouraged, we can easily reproduce the required functionality for the GOARCHs supported by pion (which would be those?).
- `RelayGenerator.AllocateConn` is back for generating the outbound TCP relay connections, so no more hardcoded `net.DialTCP` in the server. I think this is for the better: now we can abstract away all network-related functions the server invokes (e.g., to wrap OTel telemetry on top of TCP connections). This sort of fixes the API worries I mentioned in an earlier thread. 
- Unfortunately ATM we cannot abstract the TCP listener creation in `RelayGenerator.AllocateListener` via pion/transport. The reason is that `transport.Net` does not (yet?) support `net.ListenerConfig`, which we would need to enforce the reuseport semantics. Note that both `AllocateListener` and `AllocateConn` must enforce reuseport, otherwise either of them will fail with "Address already in use". Currently we hard-code the `net.ListenerConfig` in `AllocateListener` instead of going through `transport.Net`, which is on the one hand ugly and on the other hand requires adding an even uglier `context.TODO()` in `Listen`. We can easily implement something like `CreateListenerConfig` in `transport.Net` (just like we already have `CreateDialer`) and hope someone jumps in to implement TCP in pion/transport, let me know if you want be to send a PR.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -430,6 +430,9 @@ func TestTCPClientDial(t *testing.T) {
 		inboundTCPConn, inboundErr := remotePeerConn.Accept()
 		assert.NoError(t, inboundErr)
 
+		remoteAddr := inboundTCPConn.RemoteAddr()
+		assert.Equal(t, allocation.Addr().String(), remoteAddr.String(), "peer conn: remote address = relay address")
+
 		_, inboundErr = inboundTCPConn.Write(expectedMsg)
 		assert.NoError(t, inboundErr)
 
@@ -449,6 +452,9 @@ func TestTCPClientDial(t *testing.T) {
 
 	connectionBindConn, err := allocation.Dial("tcp4", remotePeerAddr.String())
 	assert.NoError(t, err)
+
+	localAddr := connectionBindConn.LocalAddr()
+	assert.Equal(t, allocation.Addr().String(), localAddr.String(), "client conn: local address = relay address")
 
 	channelBindConnBuffer := make([]byte, len(expectedMsg))
 	_, err = connectionBindConn.Read(channelBindConnBuffer)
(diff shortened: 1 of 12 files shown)


Sean-Der:
Damn this is amazing @rg0now thanks for doing this.

I would like to keep the package zero dependency. Maybe put reuse functionality in `pion/transport`? I have been burned by 3rd parties with Pion a few times :(

> We can easily implement something like CreateListenerConfig in transport.Net (just like we already have CreateDialer) and hope someone jumps in to implement TCP in pion/transport, let me know if you want be to send a PR.

Let's do that! You have my blanket support to do w/e is the right thing, I am here to approve/merge quick :)

rg0now:
See https://github.com/pion/transport/pull/372 and https://github.com/pion/transport/pull/373. Please merge and release pion/transport so that I can update this PR.

rg0now:
Thanks @JoTurk for releasing pion/transport/v4, it did not occur to me how much work this is. Sorry for making you undergo this...

Removed external dependency, restored the servers' TCP relay listener to use VNet, and fixed the StaticIP -> StaticIPs transition in pion/transport. We still indirectly depend on pion/transport/v3 (I guess via pion/stun), we may want to wait until new versions are released or merge this as is and do the version updates later.

JoTurk:
> Thanks @JoTurk for releasing pion/transport/v4, it did not occur to me how much work this is. Sorry for making you undergo this...


npnp, I wanted to do this for months since we started working on pion/SCP and was just waiting for an opportunity :)

JoTurk:
> we may want to wait until new versions are released or merge this as is and do the version updates later.

I think we should merge, we already merged breaking changes to this library.

Sean-Der:
Merge away @rg0now sorry about the merge conflicts :( 

Can I get your take on https://github.com/pion/turn/pull/464 ?

Getting very close to new release!

PR #518, opened 2026-01-07
Add Auth Check in handleConnectRequest

Sean-Der:


rg0now:
How did I not catch this during code review?? My bad... Thanks!

PR #524, opened 2026-01-10
Add `simple-quota` example

Sean-Der:
Demonstrates using the Quota + Event handlers


Sean-Der:
Here is example @rg0now. Change however you think it can be made better!

PR #526, opened 2026-01-10
[RFC] Simplified Client API

rg0now:
With the upcoming addition of TURN/TCP allocations, the mental complexity of working with the client API is becoming unwieldy. As pion/turn gains adoption in areas beyond WebRTC (VPNs, P2P networks, tunneling), this complexity may become a barrier to entry. This PR proposes a simplified client API that mimics the familiar `Dialer` and `Listener` interfaces, even at the cost of hiding some TURN complexity. For simple use cases this API should be easier to pick up, while the current client API remains available for users that want full control.

This is just a humble PR to raise some discussion, definitely not production ready.

## The Problem

With TCP allocations, we now have two fundamentally different workflows for connecting to peers:

- **UDP**: Call `client.Allocate()` -> returns `net.PacketConn` -> use `WriteTo`/`ReadFrom`
- **TCP**: Call `client.AllocateTCP()` -> returns `*client.TCPAllocation` -> call `DialTCP()` -> returns `net.Conn`

Likewise for "accepting" peer connections:

- **UDP**: `ReadFrom` on the `net.PacketConn` returned by `client.Allocate()`
- **TCP**: `Accept` on the `*client.TCPAllocation` returned from `client.AllocateTCP()`

Plus there's the complexity of managing peer permissions and `client.Listen()` that users often forget to call.

## Proposal

A unified `Allocation` interface that works like a combined `Dialer` + `Listener`:

```go
type Allocation interface {
    Dial(network, address string) (net.Conn, error)
    Accept() (net.Conn, error)
    Addr() net.Addr
    Close() error
    CreatePermission(addr net.Addr) error
}
```

Key simplifications:
- Single constructor (`NewAllocation`) for both UDP and TCP
- `Dial()` handles permission creation internally
- `Accept()` works uniformly across both transport types
- Returns standard `net.Conn` instead of protocol-specific types

### Create UDP allocation:
```
alloc, _ := turn.NewAllocation("udp", &turn.ClientConfig{
    Conn:           turnConn,
    TURNServerAddr: "turn.example.com:3478",
    Username:       "user",
    Password:       "pass",
    Realm:          "example.com",
})
defer alloc.Close()
```

For TCP, just set the network type to "tcp" and pass in a TCP TURN socket.

### Dial a peer

`Dial` creates permission automatically and returns a `net.Conn` bound to the peer both for UDP and TCP.
```
conn, _ := alloc.Dial("udp", "192.0.2.1:5000")
conn.Write([]byte("hello"))
```

### Accept incoming connections from permitted peers

Note that `Dial` automatically calls `CreatePermission`, but for listeners `CreatePermission` must be called explicitly.
```
_ = alloc.CreatePermission(peer)
incoming, _ := alloc.Accept()
// handle incoming...
```

### DIFF
--- a/allocation.go
+++ b/allocation.go
@@ -1,0 +1,385 @@
+// SPDX-FileCopyrightText: 2023 The Pion community <https://pion.ly>
+// SPDX-License-Identifier: MIT
+
+// Package turn contains the public API for pion/turn, a toolkit for building TURN clients and servers.
+package turn
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"sync"
+	"time"
+
+	"github.com/pion/transport/v4"
+	"github.com/pion/transport/v4/deadline"
+	"github.com/pion/transport/v4/packetio"
+	"github.com/pion/turn/v5/internal/client"
+)
+
+const defaultAcceptBacklog = 128
+
+var (
+	_ transport.Dialer = &turnAllocation{}
+	_ net.Listener     = &turnAllocation{}
+	_ Allocation       = &turnAllocation{}
+)
+
+// Typed errors.
+var (
+	ErrAllocationClosed = errors.New("allocation: closed")
+	errFailedToAlloc    = errors.New("allocation: failed to create allocation")
+	errFailedToDial     = errors.New("allocation: failed to dial")
+	errInvalidNetwork   = errors.New("allocation: invalid network")
+)
+
+// Allocation represents a TURN allocation that provides bidirectional communication with peers
+// through a TURN relay server. It combines dialer and listener semantics: use Dial to establish
+// outgoing connections to peers, and Accept to receive incoming connections from peers. The
+// allocation is bound to a specific transport protocol (UDP or TCP) specified when created via
+// NewAllocation. All connections through this allocation use that protocol.
+type Allocation interface {
+	// Dial establishes a connection to the specified peer address through the TURN relay.
+	// The network parameter must match the allocation's protocol ("udp" or "tcp").
+	// Dial automatically creates a TURN permission for the peer, allowing the peer to
+	// send data back to this allocation's relay address.
+	Dial(network, address string) (net.Conn, error)
+
+	// Accept waits for the next incoming connection from a peer.  The peer must have a valid
+	// permission (created via Dial or CreatePermission) to send data to this allocation's
+	// relay address.
+	Accept() (net.Conn, error)
+
+	// Addr returns the relay transport address of the allocation.
+	Addr() net.Addr
+
+	// Close releases the TURN allocation and closes all associated connections.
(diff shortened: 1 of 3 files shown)


Sean-Der:
I am all for it @rg0now as long as we drop the previous API!

* Could we be making a mistake doing UDP traffic over a `net.Conn`? If each `Read`/`Write` is one datagram seems ok?
* Does `CreatePermission` need to exist? Could `Accept(addr net.Addr)` be a thing? Would I ever need to a create a Permission if I didn't intend to accept traffic upon it? If a user wants to Accept anything pass a `nil` address?

@jech @JoTurk what's your take?

rg0now:
> I am all for it @rg0now as long as we drop the previous API!

Huhh, that's a tough call. I'm afraid this would be a massive rewrite burden on people using pion/turn. Plus the this PR merely calls into the existing API so we cannot remove it right now anyway. And then there's the issue that this API hides some features (like `PerformTransaction`, `HandleInbound`, `Connect` and `BindConnection`, etc.) that people may rely on. Plus the code should really really get some testing...

> Could we be making a mistake doing UDP traffic over a net.Conn? If each Read/Write is one datagram seems ok?

The `net.Conn` that `Dial("udp", peer)` returns is merely a wrapper on top of the existing `UDPConn`, just with binding automatically calling `WriteTo`/`ReadFrom` with `peer` as the destination address. So everything that applies to `UDPConn` applies to our `net.Conn` too.

> Does CreatePermission need to exist? Could Accept(addr net.Addr) be a thing? Would I ever need to a create a Permission if I didn't intend to accept traffic upon it? If a user wants to Accept anything pass a nil address?

I was hesitating whether to do this but eventually decided to go with `CreatePermission`/`Accept`. The problem is that there can be more than one peer that may request a connection over the allocation at any instance of time and it would be difficult to orchestrate multiple parallel blocking `Accept(peer)` calls.

Note that you only need to call `CreatePermission` if you call `Accept` before/without a `Dial` (`Dial` automatically manages permissions), but I think this is the rare case for TURN. The more I think about it the more I feel this is the right thing to do not just for TURN but maybe also for plain old UDP: it is a natural thing to want to decide who can send on a connectionless channel.

allocation.go
```diff
@@ -0,0 +1,377 @@
...
+	conn, ok := d.conns[raddr.String()]
+	if !ok {
+		// New peer - create conn and send to accept channel.
+		conn = d.newConn(raddr)
+		d.conns[raddr.String()] = conn
+		select {
+		case d.acceptCh <- conn:
+			// Sent to accept queue.
+		default:
+			// Accept queue full - conn is still registered, just not queued.
```
JoTurk: How do we handle this? do we enqueue it somewhere else? i couldn't find it.

JoTurk (commented):
I agree with the permissions API comment, maybe we just need a higher-level API?

Thank you, this is a really cool API.

Issue #364, opened 2024-01-02
integrity check failed


shanjunmei:
### Your environment.
  - Version: *Release or SHA*
  - Browser: *include version*
  - Other Information - *stacktraces, related issues, suggestions how to fix, links for us to have context*

### What did you do?
I created a service according to the code in the example, and always got the result that the integrity check failed.
`var users = "user=pass"
	usersMap := map[string][]byte{}
	for _, kv := range regexp.MustCompile(`(\w+)=(\w+)`).FindAllStringSubmatch(users, -1) {
		usersMap[kv[1]] = turn.GenerateAuthKey(kv[1], realm, kv[2])
	}
	var authHandler = func(username string, realm string, srcAddr net.Addr) ([]byte, bool) {
		if key, ok := usersMap[username]; ok {
			return key, true
		}
		return nil, false
	}
s, err := turn.NewServer(turn.ServerConfig{
		Realm: realm,
		// Set AuthHandler callback
		// This is called every time a user tries to authenticate with the TURN server
		// Return the key for that user, or false when no user is found
		AuthHandler: authHandler,
		// PacketConnConfigs is a list of UDP Listeners and the configuration around them
		PacketConnConfigs: []turn.PacketConnConfig{
			{
				PacketConn: udpListener,
				RelayAddressGenerator: &turn.RelayAddressGeneratorStatic{
					RelayAddress: net.ParseIP(publicIP), // Claim that we are listening on IP passed by user (This should be your Public IP)
					Address:      "0.0.0.0",             // But actually be listening on every interface
					//	MinPort:      32355,
					//	MaxPort:      65535,
				},
			},
		}
`

### What did you expect?
Hope it works as expected
### What happened?
turn ERROR: 2024/01/02 08:53:43 Failed to handle datagram: failed to handle Allocate-request from 8.244.68.36:55549: integrity check failed

rg0now:
I'm afraid we need more info. Which client did you use? How did you specify the username/password on the client side? Did you try the example [server](https://github.com/pion/turn/tree/master/examples/turn-server/simple) and [client](https://github.com/pion/turn/tree/master/examples/turn-client/udp) code? (This works like charm on my side.)

My first guess would be that you're not setting the client side credentials correctly, but this is just a guess. (Tracking this down would be much easier if our example codes contained proper logging, which they do not.)

shanjunmei:
[...]

PR #527, opened 2026-01-12
Fix: Example Servers Users Parsing

adrianosela:
## Fix: Example Servers Users Parsing

Fixes https://github.com/pion/turn/issues/364

There is a bug in all server examples where usernames/passwords with non-alphanumeric characters are parsed incorrectly. We are using the regex `(\w+)=(\w+)` to parse the `-users` flag, but the `\w+` pattern only matches `[a-zA-Z0-9_]`). So for example:

  - `user-name=pass-word` --> Parses as `name:pass` (drops `user-` and `-word`)
  - `user:name=pass:word` --> Parses as `name:pass` (drops everything before colons)
  - `test@example.com=P@ssw0rd!` --> Parses as `example:P` (completely breaks)
 
This PR drops the regex based logic in favor of `strings.Split`

### DIFF
--- a/examples/turn-server/add-software-attribute/main.go
+++ b/examples/turn-server/add-software-attribute/main.go
@@ -11,8 +11,8 @@ import (
 	"net"
 	"os"
 	"os/signal"
-	"regexp"
 	"strconv"
+	"strings"
 	"syscall"
 
 	"github.com/pion/stun/v3"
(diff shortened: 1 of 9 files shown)


Sean-Der:
@adrianosela amazing catch!

Should we add a `} else { panic` when things fail to split? Otherwise users might be surprised.

Would you mind squashing + adding more details to your commit message? Just copying out of GitHub 
would be perfect. Your explanation is great, don't want it to get lost :)

Merge w/e you want! Y

Issue #212, opened 2021-04-15
Example using ipv6?


morajabi:
## Summary

One paragraph explanation of the feature.

## Motivation

Why are we doing this? What use cases does it support? What is the expected outcome? Is this available in other implementations? Can this not be implemented using primitives?

## Describe alternatives you've considered

A clear and concise description of the alternative solutions you've considered.

bekriebel:
+1 to this because I'm also trying to use fly.io.

In particular, I'd love to be able to have TURN listening on ipv4 and ipv6 UDP and TCP and have it relay for an SFU running on ipv6 UDP.

GingerAdonis:
Unfortunately it requires the implementation of RFC 6156 to support IPv6.

GingerAdonis:
@Sean-Der Is IPv6 support planned? Thanks in advance.

stv0g:
Seems like the TURN client currently does IPv4 lookups only:

https://github.com/pion/turn/blob/6880381c77d370eb8121fe7eb237b838ea8e4429/client.go#L95

stv0g:
See also #265

PR #528, opened 2026-01-12
RFC 6156: TURN Extension for IPv6

adrianosela:
## RFC 6156: TURN Extension for IPv6

Closes https://github.com/pion/turn/issues/212

This PR implements [RFC 6156: Traversal Using Relays around NAT (TURN) Extension for IPv6](https://www.rfc-editor.org/rfc/rfc6156) enabling UDP and TCP relay allocations on IPv6 addresses.

cc:// @stv0g, @Sean-Der

## Testing:

- Added relevant unit tests
- Added client-server example:

```
✔ ~/go/src/github.com/adrianosela/turn [rfc6156-ipv6-allocs|⚑ 2]
19:21 $ go run examples/turn-server/ipv6/main.go -public-ip ::1 -users username=password
2026/01/11 19:21:34 TURN server listening on [::]:3478 with IPv6 support (RFC 6156)
2026/01/11 19:21:34 Public IPv6 address: ::1
```

```
19:21 $ go run examples/turn-client/ipv6/main.go -host ::1 -user username=password -ping
2026/01/11 19:21:58 relayed-address=[::1]:61557 (IPv6)
2026/01/11 19:21:58 mapped-address=[::1]:52873
2026/01/11 19:21:59 32 bytes from from [::1]:61557 time=1 ms
2026/01/11 19:22:00 32 bytes from from [::1]:61557 time=1 ms
2026/01/11 19:22:01 32 bytes from from [::1]:61557 time=1 ms
2026/01/11 19:22:02 32 bytes from from [::1]:61557 time=2 ms
2026/01/11 19:22:03 32 bytes from from [::1]:61557 time=1 ms
```


Sean-Der:
Hot damn that's amazing!

Mind squashing into one commit and copying your GitHub message into the commit? Reviewing now.

README.md
```diff
@@ -66,13 +66,17 @@ and matches the certificate's Common Name to the TURN username.
 #### Will WebRTC prioritize using STUN over TURN?
 Yes.
 
+#### How do I use IPv6 with TURN?
```
Sean-Der: What do you think about inferring what you use from what was passed by the user?

Pull the `LocalAddr` and check for IPv4 and IPv6 support?

Sean-Der:
@adrianosela LGTM! This is really impressive stuff, I am so appreciate of you getting involved :)

Sean-Der:
Approved, see if you can do the following (totally understand if it isn't possible)

* Can we choose the transport from looking at passed PacketConn?
* Fix up linters
* Fix up commit message
* Possible to raise the code coverage a bit more? Error handling in `server/turn.go` would be nice

adrianosela:
@Sean-Der , all reasonable feedback. Will get to it over next few days/weeks.

adrianosela:
@Sean-Der 

- ✅ Linters
- ✅ Commit message
- ✅ Raise coverage

> * Can we choose the transport from looking at passed PacketConn?

So, server side we have to rely on the `REQUESTED-ADDRESS-FAMILY` as per the RFC. But on the client side we can *try* to infer it from the packet conn, but we cannot get rid of the additional `RequestedAddressFamily` field because the address methods of `net.PacketConn` are not necessarily IPs.

They will be for most, so we can definitely *try*. I propose we:

  1. Make RequestedAddressFamily optional (and we treat the zero value as not set)
  2. If set, use it
  3. If not set, try infer from config.Conn.LocalAddr()
  4. Default to IPv4 if inference fails

I'll make that change and ask you for another review.

client.go
```diff
@@ -126,7 +149,7 @@ func NewClient(config *ClientConfig) (*Client, error) {
 	}
 
 	if len(config.TURNServerAddr) > 0 {
-		turnServ, err = config.Net.ResolveUDPAddr("udp4", config.TURNServerAddr)
+		turnServ, err = config.Net.ResolveUDPAddr(network, config.TURNServerAddr)
```
JoTurk: I'm going off Sean's comment here and a basic reading of the spec but isn't REQUESTED-ADDRESS-FAMILY is only about the relayed address family the server allocates. It does not suggest that the client must contact the server over IPv6. am i missing something?

adrianosela:
Okay, this is ready for another review.  cc:// @JoTurk @Sean-Der

PR #529, opened 2026-01-13
Consistent Joining of Host + Port

adrianosela:
## Consistent Joining of Host + Port

Now that RFC 6156: TURN Extension for IPv6 is implemented, we should get used to using `net.JoinHostPort` which is the canonical way to join a host and port (handling the v4 and v6 cases).

This PR also standardizes on `strconv.Itoa` (instead of `fmt.Sprintf` for integer-to-string conversions of ports).


adrianosela:
The changes in `internal/proto/addr.go` and `relay_address_generator_none.go` are bug fixes too.

Sean-Der (approved):

Issue #150, opened 2020-08-01
Track amount of bandwidth utilized overall and by a specific user.


jipson7:
## Summary

Would like to know how to track the bandwidth used by a user on the turn server. Or simply overall bandwidth to start.

## Motivation

Primarily for billing and book keeping.

## Describe alternatives you've considered

I looked at the turn-server logging example but couldn't see anything obvious that would lend itself to tracking bandwidth.

Sean-Der:
@rg0now @adrianosela What do you think about this one?

An API that allows me to know

* Usage per Allocation
* Usage to Destination
* usage from Destination
* ?

Adding something like this to EventHandler?

```
      - OnTraffic func(srcAddr, dstAddr net.Addr, protocol, username, realm string, relayAddr, peerAddr net.Addr, direction TrafficDirection, bytes int)
      - TrafficDirection enum: ClientToPeer, PeerToClient
```

adrianosela:
@Sean-Der , do we want to add to the api the ability to drop the traffic too? if it exceeds bandwidth limits for example. Or is there already another mechanism to do something like that?

Sean-Der:
None exists! I am torn between the following

* Expose a `Allocation` object?
* Add APIs like `CloseAllocation` on the server?

What feels right to me is `GetAllocation(FiveTuple)` on the server. We start really simple and just have this?

```
struct Allocation {}

func (a *Allocation) Close() error {} 
```

We can add more in the future

@rg0now ?

Sean-Der:
maybe we save for next release @adrianosela I really would like to have you announce a new release in next day or so.

As long as we can add this other stuff as non-breaking changes, we should hold off! We can do an exciting v5.1 in a month.

rg0now:
I think the right way to do this is to gently regulating users' traffic to remain under the quota instead of abruptly terminating allocations exceeding it (which one? a user may have tons of allocations and bw quotas apply to the total traffic). The way I'd implement this is wrapping the connections returned from `RelayAddressGenerator.AllocatePacketConn` and `RelayAddressGenerator.AllocateConn` with a rate-limiter that consumes tokens from a per-user token bucket and drops packets that exceed the quota. (Do we need 2 token buckets, one for the `Read` and one for the `Write` direction??) I'll try to contribute an example.

[...]

PR #531, opened 2026-01-13
Make RelayGen allocator API extensible

rg0now:
See #150 and [this comment](https://github.com/pion/turn/issues/150#issuecomment-3743797983)

There has been a couple of requests to extend the RelayGenerator listener/connection allocation API with new parameters. Especially adding the username/realm to the allocators' parameter list is compelling, in order to "personalize" relay address/connection allocations and implement per-user bandwidth quotas. This PR aims to make the API extensible by packing all parameters into a struct.

### DIFF
--- a/internal/allocation/allocation_manager.go
+++ b/internal/allocation/allocation_manager.go
@@ -20,12 +20,53 @@ import (
 // MUST be closed.
 const defaultTCPConnectionBindTimeout = time.Second * 30
 
+// AllocateListenerConfig contains the parameters passed to the relay address allocator
+// when creating a new UDP or TCP allocation.
+type AllocateListenerConfig struct {
+	// Network specifies the network type for the allocation: "udp4", "udp6", "tcp4", or "tcp6".
+	Network string
+	// UserID is the authenticated user's identifier as returned by the AuthHandler.
+	//
+	// Note: The UserID is typcally the same as the TURN username, except for authentication
+	// schemes that overload the username field with additional info (e.g., the lifetime of the
+	// credential, as in the time-windowed credential mechanism in
+	// https://datatracker.ietf.org/doc/html/draft-uberti-behave-turn-rest-00.
+	UserID string
+	// Realm is the TURN realm for this allocation.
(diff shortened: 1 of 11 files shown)


Sean-Der:
Nice LGTM @rg0now 

We should add examples of Quotas + Tracking + Rate Limiting so we can resolves issues. 

Great call out/design on sticking to the Conns :)

rg0now:
> We should add examples of Quotas + Tracking + Rate Limiting so we can resolves issues.

See #532 

It just occurred to me that there is a fundamental catch here: in the example we apply rate-limiting *after* we processed the packet via the TURN server. I don't see any way to add the rate-limiter *before* the TURN server processing though because we need to identify the username but by the time we get there we've already done a fair amount of processing. 

As per the original request: the same idea can be used to track user traffic bw instead of rate-limiting it.

rg0now:
I accidentally closed this... Reopening now

internal/allocation/allocation_manager.go
```diff
@@ -20,12 +20,27 @@ import (
 // MUST be closed.
 const defaultTCPConnectionBindTimeout = time.Second * 30
 
+type AllocateListenerInfo struct {
+	Network       string
+	Username      string
+	Realm         string
```
JoTurk: I commented on the original issue https://github.com/pion/turn/issues/150#issuecomment-3746604173 Please don't treat this as a blocker. I'm just sharing my instinct.

I think the authentication layer should return an application-level id (it can be anything depending on the application), and it shouldn't leak authentication info. Otherwise it feels like we're building two authentication layers.

For example, an application should be able to map authentication into tiers or rate-limit buckets. With this design, I think applications will have to make the allocation layer auth-aware.

JoTurk (approved):

internal/allocation/allocation_manager.go
rg0now: @JoTurk See #538

rg0now:
I braved up and merged this one too. What a day!

Issue #229, opened 2021-08-20
Forcing relay only on my clients and a port allocator results in never allocating a port


doctorpangloss:
### Your environment.
  - Version: 72ffd594f7c6104b5680a4fea5651928e3b6ddb4
  - Browser: Mobile Safari on iOS 12.7, Chrome 92

### What did you do?
I added logging to `RelayAddressGeneratorPortRange` to see if a port is ever allocated.

```go
	for try := 0; try < r.MaxRetries; try++ {
		port := r.MinPort + uint16(r.Rand.Intn(int((r.MaxPort+1)-r.MinPort)))
		conn, err := r.Net.ListenPacket(network, fmt.Sprintf("%s:%d", r.Address, port))
		if err != nil {
		  log.Printf("could not allocate port %d", port)
			continue
		}

		relayAddr := conn.LocalAddr().(*net.UDPAddr)
		relayAddr.IP = r.RelayAddress
		log.Printf("allocated %d")
		return conn, relayAddr, nil
	}
```

### What did you expect?

Ports should be allocated.

### What happened?

A port is never allocated randomly within the range. I see multiple of the following errors:

```
turn ERROR: 2021/08/19 20:17:08 error when handling datagram: failed to handle Allocate-request from 192.168.88.162:61561: relay already allocated for 5-TUPLE
```

adrianosela:
@doctorpangloss thank you for reporting.

I'm digging into this.

adrianosela:
So, I'm not sure if this is the bug the reporter experienced or even a bug at all, but I was looking into the error message.

In `internal/server/turn.go`:

```
if alloc := req.AllocationManager.GetAllocation(fiveTuple); alloc != nil {
    id, attrs := alloc.GetResponseCache()
    if id != stunMsg.TransactionID {
        // reject with error
    }
}
```

**We reject allocations for 5-tuples already allocated if the transaction ID does not match the original request, even if the user is the same. I believe some clients may retry with different tx id(?) so perhaps we should consider returning successfully if the user is the same (despite the tx id mismatch). Unless this behavior is explicitly part of an RFC?**

To illustrate the point:

Say a client makes an allocation request:
- user: abc
- 5-tuple: (192.168.1.100:50000 -> 10.0.0.1:3478, UDP)
- tx id: 1111

And the allocation succeeds server side but the client thinks it failed (due to packet loss, etc), the client may retry with the same user and 5-tuple, but different tx id:
- user: abc
[...]

PR #533, opened 2026-01-14
Tests for RFC Adherence on Allocation Retries

adrianosela:
## Tests for RFC Adherence on Allocation Retries

Closes https://github.com/pion/turn/issues/229.

Test RFC 5766 Section 6.2 compliance with respect to allocation retries.

On a separate note, perhaps we should consider adding a retry mechanism to the API in a way that `ListenPacket` retries for the same 5-tuple can re-use the same Transaction ID... To account for allocations that succeeded but the response never made it to the client e.g. due to packet loss, etc.

### DIFF
--- a/internal/server/turn_test.go
+++ b/internal/server/turn_test.go
@@ -883,6 +883,112 @@ func TestHandleChannelBindRequest(t *testing.T) {
 	})
 }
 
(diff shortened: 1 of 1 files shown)


JoTurk (approved):
cool.

PR #534, opened 2026-01-14
Tighten IPv6 address-family handling

JoTurk:
#### Description

I had nothing on my todo list today so i thought with helping testing the new turn stuff \:)
Found few nits with IPV6 handling while testing on an IPv6 network.

this does:
- honor requested address family for TCP connections
- return 443 on invalid/mismatched requested-family in refresh
- preserve IPv6 zone when deriving TCP server addr.


JoTurk:
(Also I'm going to test other new stuff in pion/turn tomorrow, many cool new stuff :)

adrianosela (approved):
🚢 it, and thank you!

PR #536, opened 2026-01-14
Fix TestCreateTCPConnection accept race

JoTurk:
#### Description
Wait for the accept goroutines to finish, fixes a panic in test, happens sometimes when I run the test locally and happens in the CI too https://github.com/pion/turn/actions/runs/20979502198/job/60301315924


adrianosela (approved):
:shipit:

PR #532, opened 2026-01-13
Bandwidth quota example

rg0now:
See #150 

This demo shows how to apply per-user rate-limiting at the peer connection by maintaining a pool of per-user rate-limiters and using a custom `RelayAddressGenerator` that generates rate-limited relay-connections.

PR #538, opened 2026-01-14
Return the parsed user-id from the AuthHandler

rg0now:
This PR changes the `AuthHandler` to return the user-id from the auth layer. The user-id is typically the same as the TURN username, except for authentication schemes that overload the username with additional info (like [time-windowed TURN credentials](https://datatracker.ietf.org/doc/html/draft-uberti-behave-turn-rest-00)). The rest of the server stack then uses the user-id instead of the username to identify users in the various handlers, which makes sure that, e.g., the `EventHandler` API will identify allocations per user-id and the same user-id is then passed to the `QuotaHanlder` and the `RelayAddressGenerator` socket allocators. This prevents errors when, e.g., we account two allocations created with the same user-id but with different timestamps as two different users in the quota handler.

Merging strategy: if this is merged, then rebase and merge #531, and then rebase and merge #532.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -153,8 +153,8 @@ func TestClientNonceExpiration(t *testing.T) {
 	assert.NoError(t, err)
 
 	server, err := NewServer(ServerConfig{
-		AuthHandler: func(ra *RequestAttributes) (key []byte, ok bool) {
-			return GenerateAuthKey(ra.Username, ra.Realm, "pass"), true
+		AuthHandler: func(ra *RequestAttributes) (userID string, key []byte, ok bool) {
+			return ra.Username, GenerateAuthKey(ra.Username, ra.Realm, "pass"), true
 		},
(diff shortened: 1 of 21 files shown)


rg0now:
> can UserID be empty? when ok is true? I think we should be explicit about with unit test, whether we allow or disallow empty userids.

I think it can. Added a test and it passes...:-) Also updated the quota handler test to make sure we pass the user-id, not the username.

rg0now:
@JoTurk @Sean-Der Should we go on with the plan (merge this, then rebase and merge #531 and #532, in that order)? Do you want me to do the merge?

adrianosela (approved):

rg0now:
@Sean-Der I went ahead and merged this, pls revert if you don't agree.

Sean-Der:
@rg0now I love when you merge things! I think your opinion is better/worth more here.

you have a active product, use this project everyday :) so I’m here to support you!

PR #539, opened 2026-01-15
Fix UDPConn.WriteTo to return the payload length

rg0now:
`UDPConn.WriteTo` has two paths: write via ChannelData (after binding channel) or write via SendIndication (before binding). The SendIndication path was returning the STUN message size instead of the actual payload size (the ChanelData write-path is already returning the correct payload length). This breaks `io.Copy` which expects Write to return exactly the number of bytes from the source buffer. When this invariant does not hold, io.Copy fails with "invalid write result". 

This fix ensures both paths return len(payload) per the net.PacketConn contract.


adrianosela (approved):

JoTurk:
nice  :)

PR #546, opened 2026-01-27
Support alternate for UDP allocation

### LINKED ISSUE #544: Error code 300 gets ignored
Hi, If a TURN server responds with Allocate error response (error 300: Try Alternate) then webrtc will fail to get a relay ip causing relay only mode to break and fallback connection.

ice WARNING: 2026/01/02 06:11:40 failed to allocate on TURN client relay.communication.microsoft.com:3478 Allocate error response (error 300: Try Alternate)

JoTurk:
#### Description

This is a draft for supporting alternate addresses, this works for UDP.

two things:
1. to support alternate for TCP allocation we'll have to add a connection-factory helper or something so users can provide connections for alternate addresses or create the connections our-self, i feel like v5 was a missed opportunity to fix this. 
2. I reviewed the libwebrtc code and a few other libraries and couldn't find any explicit limit on the number of alternatives they follow. idk that felt risky to me, so I hard-coded a limit of 64, which I think is very generous.

#### Reference issue

Fixes #544

### DIFF
--- a/client.go
+++ b/client.go
@@ -5,6 +5,7 @@ package turn
 
 import (
 	b64 "encoding/base64"
+	"errors"
 	"fmt"
 	"math"
 	"net"
@@ -23,6 +24,10 @@ const (
 	defaultRTO        = 200 * time.Millisecond
 	maxRtxCount       = 7              // Total 7 requests (Rc)
 	maxDataBufferSize = math.MaxUint16 // Message size limit for Chromium
+	// while there is no limit in the RFC (RFC 5389/8489),
+	// this caps how many consecutive ALTERNATE-SERVER redirects
+	// we have a generous limit of 64 to prevent infinite loops and abuse.
+	maxAltSrvRedirects = 64
 )
 
 //              interval [msec]
@@ -67,7 +72,7 @@ type Client struct {
 	conn           net.PacketConn // Read-only
 	net            transport.Net  // Read-only
 	stunServerAddr net.Addr       // Read-only
-	turnServerAddr net.Addr       // Read-only
+	turnServerAddr net.Addr
 
 	username      stun.Username          // Read-only
 	password      string                 // Read-only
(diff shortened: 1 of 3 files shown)


JoTurk:
@sean-bauer @rg0now @adrianosela I'm going to keep this a draft until we decide what to do with TCP, I think our options are: either add a dial/connection-factory in config so AllocateTCP() can re-connect, or keep this UDP-only for now and have TCP either fail on 300 or return the alternate addr so callers can retry with a new client.

rg0now:
Interesting. Kept me a while to think through.

One option is to change the `turn.Client` to receive a `net.Dialer` instead of a `net.PacketConn` (this would be your connection factory?) and then let `Allocate` create the server connections for itself and encapsulate the retry logic. The `net.Dialer` could enforce the redirect limit by failing to create new connections over a certain quota, or we could make the quota explicit in the config. If the caller wants glorified sockets (e.g., enriched with Prom counters), they'd need to pass us a `net.Dialer` that generates such connections. This covers both TCP and UDP. The cost is that we would have to break the API (again) and change the entire client logic: the client now expects a `net.PacketConn` and uses `WriteTo/ReadFrom` to send to/receive from the server. However, `net.Dialer` produces a `net.Conn` that is already connected to the server so internally we should use `Write/Read` for sending/receiving stuff. I fail to see why this would go wrong; in fact a connected socket feels more correct since we always speak to the same TURN server (minus `TryAlternate`). 

The second option is to rely on the caller to track the alternates and implement the retry logic. I mean we may want to fail on 300 AND return the alternate addr in the err so callers can retry with a new client. This looks simpler, at the cost of the caller having to create multiple clients/connections until they succeed (which sux). I don't think we can make the same client reusable for multiple `client.Allocate` calls, or at least I think we cannot reliably reuse a TCP server connection that we get from the user.

The third option is implement the redirect logic for UDP and fail on 300 for TCP.  I guess this is what you're doing here, right?

JoTurk:
I think we'll have to split this into two:

1. New API with dial-factory or Dialer, we support TCP alternate.
2. Old API, only udp alternate and we return error for tcp.

I think this makes sense. 
sorry for spamming @ these days @rg0now @Sean-Der @adrianosela

JoTurk:
(sorry for ghosting this, I didn't see your response).

rg0now:
@JoTurk The more I think of this the more I feel we should not hardcode the retry logic in the client and we should rely on the caller to implement the retry logic themselves. What if they have to use a new auth credential for the alternate? Maybe they have to make an external call to grab the new time-windowed credential? Or log/authorize TURN alternates locally?

I think we should expose your `errAlternateServerError` and return that to the client on redirect, and provide an example on how to easily track alternates. Wdyt?

JoTurk:
@rg0now yeah this is a great idea, also with this i think we can do it without breaking the API for TCP, I'll update the PR.

Issue #566, opened 2026-05-28
RFC 6156 default-IPv4 family blocks TURN on IPv6-only Kubernetes clusters


ti:
## Summary

`internal/server/turn.go ipMatchesFamily` rejects pure-IPv6 peer
addresses against the RFC 6156 default IPv4 allocation family that
pion/turn applies when an `Allocate` request omits
`REQUESTED-ADDRESS-FAMILY`. Combined with browsers always omitting
that attribute, this makes pion/turn unusable as a TURN relay in front
of IPv6-only or single-stack-IPv6 backends — which is the standard
networking model for several Kubernetes flavours, including offerings
that default to IPv6-only pod CIDRs.

## Repro

IPv6-only Kubernetes cluster, all pod IPs are pure IPv6. Backend
media-server pod (a pion-based WebRTC SFU) listens on its IPv6 pod
address on a UDP port. Browser uses pion/turn-based TURN server with
`iceTransportPolicy: "relay"`; the Allocate omits
`REQUESTED-ADDRESS-FAMILY` (default IPv4 per RFC 6156):

```
Allocate ACCEPTED
Allocation created, relay-address=<v6-relay>:<relay-port>
CreatePermission ACCEPTED  (auth)            ← OnAuth fires
                                              ← but ipMatchesFamily silently rejects
                                              ← addPermission never runs
Send-indication                              ← dropped: "no permission added"
```

The `peer address family mismatch` log line in
`internal/server/turn.go:398,511` appears at `Info` level only, easy
to miss; the `Adding permission for …` debug line never fires; the
client gets a CreatePermission error response and the call times out.

End-to-end traces on browsers (Chrome, Firefox, Safari): none of them
ever attach `REQUESTED-ADDRESS-FAMILY` to the Allocate request, so
every browser hits this path on an IPv6-only cluster.

## Root cause

`internal/server/turn.go` line 174:

```go
case errors.Is(err, stun.ErrAttributeNotFound):
    requestedFamily = proto.RequestedFamilyIPv4   // RFC 6156 default
```

Per [RFC 6156 §4.2.2.1](https://datatracker.ietf.org/doc/html/rfc6156#section-4.2.2.1)
the default is "IPv4" so the allocation is IPv4-family. Then at line
396 (CreatePermission) and 509 (ChannelBind):

```go
if !ipMatchesFamily(peerAddress.IP, alloc.AddressFamily()) {
    req.Log.Infof("peer address family mismatch …")
[...]

PR #567, opened 2026-05-28
fix(server): accept v6 peers on default-IPv4 allocations for dual-stack relays

ti:
## Problem

`internal/server/turn.go ipMatchesFamily` strictly rejects pure-IPv6
peer addresses against an IPv4 allocation, including the **default**
IPv4 allocation that pion/turn applies when an `Allocate` request
omits `REQUESTED-ADDRESS-FAMILY` (RFC 6156 §4.1.1).

Browser WebRTC TURN clients (Chrome, Firefox, Safari) **never** set
`REQUESTED-ADDRESS-FAMILY` in their `Allocate` requests, so on a
dual-stack or IPv6-only Linux host every browser-driven allocation
ends up IPv4-family even when the listener and relay sockets are
bound `[::]`. Subsequent `CreatePermission`/`ChannelBind` requests
with v6 peer addresses are then rejected with
`errPeerAddressFamilyMismatch` (logged at `Info` level, easy to
miss), and the very first `Send-indication` fails with
`errNoPermission`.

This is the canonical failure mode for pion/turn-based TURN servers
deployed in front of IPv6-only Kubernetes pod networks. With the
relay socket actually bound dual-stack on Linux (`net.ipv6.bindv6only=0`,
which is the default), v6 peer reachability is fine; only the
protocol-level family check blocks the permission.

## Repro & root-cause

Detailed write-up + minimal repro in **#566**.

Short version: with a Linux dual-stack relay socket and a v4-default
allocation, peer reachability works in both directions but
`ipMatchesFamily(v6-peer, IPv4-family)` returns false ⇒ no
permission ⇒ no Send-indication forwarded ⇒ WebRTC PC times out.

## Fix (this PR)

Relax the v4-default branch of `ipMatchesFamily` to also accept v6
peers, matching what a dual-stack relay socket can actually deliver.

```go
 if family == proto.RequestedFamilyIPv4 {
-    return ip.To4() != nil
+    return ip.To4() != nil || (ip.To4() == nil && ip.To16() != nil)
 }
```

The symmetric IPv6-allocation case is **unchanged**: a v4 peer on a
v6-explicit allocation is still rejected, preserving RFC 6156
strict-mode for the case where the client opted into v6 by sending
`REQUESTED-ADDRESS-FAMILY=IPv6`.

`Refresh` request family handling is untouched.

## Tests

`internal/server` had two assertion sites baked to the strict cross-
family rejection. They are updated to match the new permissive
behavior:

- `TestIPMatchesFamily/IPv6` and `/IPv6Loopback`: `ipMatchesFamily(v6,
  RequestedFamilyIPv4)` now returns `true`.
- `TestHandleChannelBindRequest/PeerAddressFamilyMismatch` is renamed
  to `CrossFamilyV6PeerOnV4Alloc` and asserts the success response
  (the previous request flow now produces `ClassSuccessResponse`
  instead of 443 `PeerAddrFamilyMismatch`).

All other tests in `./...` pass unchanged.

## Caveats / alternatives discussed in #566

This is the simplest of the three options I sketched in #566. The
other two would be more conservative:

1. **Default the allocation family to the listener's family** when
   `REQUESTED-ADDRESS-FAMILY` is absent — backwards-compatible for
   IPv4-only deployments, correct for IPv6-only listeners.
2. **A `Server` config knob** (`PermissiveAddressFamily bool` or
   similar) that opts in to relaxed cross-family. RFC 6156 strict
   stays the default.

I went with the simplest patch first because it fixes the canonical
deployment case with the smallest diff. Happy to refactor to either
option (1) or (2) if maintainers prefer — let me know.

## Compatibility note

This subtly changes behavior for clients that explicitly relied on
the v4-default + v6-peer = 443 response to detect IPv6 capability.
In practice no real-world TURN client does this (they all either
omit the attribute or explicitly request a family); but worth
flagging for review.

Refs: #566

### DIFF
--- a/internal/server/turn.go
+++ b/internal/server/turn.go
@@ -744,9 +744,22 @@ func handleConnectionBindRequest(req Request, stunMsg *stun.Message) error {
 }
 
 // ipMatchesFamily checks if an IP address matches the given address family.
+//
+// Relaxed for dual-stack/IPv6-only hosts: when the allocation declared the
+// RFC 6156 default IPv4 family but the relay socket was bound dual-stack
+// ("udp" wildcard) the v6 peer is reachable through the same allocation,
+// so we accept it. Strict RFC 6156 behavior is preserved for IPv6
+// allocations (a v4 peer is still rejected).
+//
+// Background: browsers' WebRTC TURN clients never set
+// REQUESTED-ADDRESS-FAMILY, so every Allocate is implicitly IPv4 family.
+// On IPv6-only Kubernetes clusters (where backend pod addresses are
+// pure IPv6, e.g. EKS Auto with IPv6 STACK_TYPE) that makes every
+// CreatePermission to a v6 peer fail with errPeerAddressFamilyMismatch
+// even though the relay socket itself is reachable from v6.
 func ipMatchesFamily(ip net.IP, family proto.RequestedAddressFamily) bool {
 	if family == proto.RequestedFamilyIPv4 {
-		return ip.To4() != nil
+		return ip.To4() != nil || (ip.To4() == nil && ip.To16() != nil)
 	}
 	if family == proto.RequestedFamilyIPv6 {
 		return ip.To4() == nil && ip.To16() != nil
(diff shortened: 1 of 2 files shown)


rg0now:
See https://github.com/pion/turn/issues/566#issuecomment-4574673996

PR #570, opened 2026-05-30
Fix PermissionHandler not enforced for TURN TCP

sirzooro:
## Summary

This PR fixes a security issue where the server-configured `PermissionHandler` was not consulted for TURN TCP connections, allowing clients to reach peer addresses that should have been blocked.

Two related paths were affected:

1. **Outbound (client-initiated) TCP connections via CONNECT** — `handleConnectRequest` in `internal/server/turn.go` called `CreateTCPConnection` (which dials the peer) without first invoking `GrantPermission`. A blocked peer could be dialled before the handler had any chance to deny it.

2. **Inbound TCP connections accepted on the relay listener** — `connHandler` in `internal/allocation/allocation.go` accepted connections from inbound peers and forwarded a `ConnectionAttempt` indication to the TURN client without checking `GrantPermission`. A peer whose IP was blocked by `PermissionHandler` could still initiate a connection that the TURN client would see.

For context: the `PermissionHandler` is already enforced for UDP (`CreatePermission`, `ChannelBind`, `SendIndication`) and was previously enforced for `ChannelBind` on TCP as well. These two TCP paths were the gap.

---

### `internal/server/turn.go`

Added a `GrantPermission` check in `handleConnectRequest` immediately after the XOR-PEER-ADDRESS is parsed, before `CreateTCPConnection` is called. If the `PermissionHandler` blocks the peer, the server returns a 403 Forbidden error response and stops processing:

```go
if err = req.AllocationManager.GrantPermission(req.SrcAddr, peerAddr.IP); err != nil {
    req.Log.Infof("permission denied for client %s to peer %s", req.SrcAddr, peerAddr.IP)

    return buildAndSendErr(req.Conn, req.SrcAddr, err, buildMsg(
        stunMsg.TransactionID,
        stun.NewType(stun.MethodConnect, stun.ClassErrorResponse),
        &stun.ErrorCodeAttribute{Code: stun.CodeForbidden},
    )...)
}
```

### `internal/allocation/allocation.go`

Added a `GrantPermission` check in `connHandler` after a new inbound TCP connection is accepted and its remote address is confirmed to be a `*net.TCPAddr`. If the `PermissionHandler` blocks the peer, the connection is closed immediately and the loop continues — no `ConnectionAttempt` indication is sent to the TURN client:

```go
if err := manager.GrantPermission(a.fiveTuple.SrcAddr, tcpAddr.IP); err != nil {
    _ = conn.Close()
    a.log.Infof("Relay %s: permission denied for inbound peer %s, closing connection",
        a.relayListener.Addr(), tcpAddr.IP)

    continue
}
```

### `internal/server/turn_test.go`

Added `TestConnectRequestPermissionDenied`: verifies that when `PermissionHandler` returns `false` for a peer IP, `handleConnectRequest` returns an error, the `PermissionHandler` is actually invoked, and `AllocateConn` (the dial) is never called.

### `internal/allocation/allocation_test.go`

Added `TestAllocationConnHandler_PermissionDeniedClosesConn`: verifies that when `PermissionHandler` returns `false` for an inbound peer, `connHandler` closes the connection, the `PermissionHandler` is actually invoked, and no `ConnectionAttempt` indication is written to the TURN socket.

### DIFF
--- a/internal/allocation/allocation.go
+++ b/internal/allocation/allocation.go
@@ -450,6 +450,14 @@ func (a *Allocation) connHandler(manager *Manager) {
 		}
 		a.log.Debugf("Relay %s accepted connection from %s", a.relayListener.Addr(), tcpAddr)
 
+		if a.GetPermission(tcpAddr) == nil {
+			_ = conn.Close()
+			a.log.Infof("Relay %s: no permission for inbound peer %s, closing connection", a.relayListener.Addr(),
+				tcpAddr.IP)
+
+			continue
+		}
+
 		cid, err := manager.addTCPConnection(a, conn)
 		if err != nil {
 			a.log.Errorf("Failed to create inbound TCP Connection for Allocation %v %v", a.fiveTuple.SrcAddr, err)
--- a/client_test.go
+++ b/client_test.go
@@ -513,6 +513,9 @@ func TestTCPClientAccept(t *testing.T) {
 	allocation, err := client.AllocateTCP()
 	assert.NoError(t, err)
 
+	// Install a permission for the loopback peer before it connects (RFC 6062 §5).
+	require.NoError(t, client.CreatePermission(&net.TCPAddr{IP: net.ParseIP("127.0.0.1")}))
+
 	expectedMsg := []byte{0xDE, 0xAD, 0xBE, 0xEF}
 
 	peerConn, err := net.Dial("tcp4", allocation.Addr().String()) // nolint: noctx
@@ -586,6 +589,9 @@ func TestTCPClientMultipleConns(t *testing.T) {
 
 	allocation, err := client.AllocateTCP()
 	assert.NoError(t, err)
+
+	// Install a permission for the loopback peer before it connects (RFC 6062 §5).
+	require.NoError(t, client.CreatePermission(&net.TCPAddr{IP: net.ParseIP("127.0.0.1")}))
 
 	var wg sync.WaitGroup
 	runPeerDialer := func(i int) net.Conn {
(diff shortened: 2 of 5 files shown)


rg0now (approved):

rg0now:
Hmm, how did we overlook this? 

@JoTurk please merge (I can also merge if you approve)

internal/allocation/allocation.go
```diff
@@ -450,6 +450,14 @@ func (a *Allocation) connHandler(manager *Manager) {
 		}
 		a.log.Debugf("Relay %s accepted connection from %s", a.relayListener.Addr(), tcpAddr)
 
+		if err = manager.GrantPermission(a.fiveTuple.SrcAddr, tcpAddr.IP); err != nil {
+			_ = conn.Close()
+			a.log.Infof("Relay %s: permission denied for inbound peer %s, closing connection", a.relayListener.Addr(),
+				tcpAddr.IP)
+
+			continue
+		}
```
JoTurk: We should also call GetPermission, no?

JoTurk:
> @JoTurk please merge (I can also merge if you approve)

I added a comment, but please feel to merge anything you see fit :)

internal/allocation/allocation.go
rg0now: Huhh, I forgot the `GetPermission` (which just checks if an already added permission exists) vs `GrantPermission` (which actually calls into the configured permission handler to check, this is to be called from `CreatePermission`) controversy. The two things should return more or less the same (unless the permission handler is stateful) but in this case it should be `GetPermission`, not `GrantPermission`

rg0now:
Your comment is actually a blocker (AFAIU)...:-)

rg0now:
Dear @sirzooro, can you jump in to fix the comment from @JoTurk ? Essentially we'll have to call `GetPermission` instead of `GrantPermission`. I'm happy to contribute a fix but we'd like the PR to be attributed to you. And thanks anyway, this is an important security fix.

sirzooro:
Yes, I am going to do it later today. Yesterday I was busy with other things :)

internal/allocation/allocation.go
sirzooro: Fixed, please take a look again.

JoTurk (approved):

JoTurk:
thank you @sirzooro

PR #575, opened 2026-06-09
Default omitted address family to listener family

### LINKED ISSUE #566: RFC 6156 default-IPv4 family blocks TURN on IPv6-only Kubernetes clusters

JoTurk:
#### Description

Browsers and some clients don't send THE RFC-6156 REQUESTED-ADDRESS-FAMILY attribute on allocate request when connecting to TURN servers over IPv6. Instead they just send RFC-5766-like allocate requests (even tho this spec doesn't support IPv6) and pion/turn ends up default to IPv4 which breaks IPV6-only listeners.

This changes the default behavior to derive the allocation family from the listener address when requested address family is omitted. Also adds a new optin config  StrictAddressFamily for users who want a strict RFC-6156 behavior for whatever reason.

I tested with chrome in a docker setup and it seems to be working.

#### Reference issue
Fixes #566

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -38,9 +38,10 @@ type Request struct {
 	Log   logging.LeveledLogger
 	Realm string
 
-	ChannelBindTimeout time.Duration
-	PermissionTimeout  time.Duration
-	AllocationLifetime time.Duration
+	ChannelBindTimeout  time.Duration
+	PermissionTimeout   time.Duration
+	AllocationLifetime  time.Duration
+	StrictAddressFamily bool
 }
 
 // HandleRequest processes the give Request.
--- a/internal/server/turn.go
+++ b/internal/server/turn.go
@@ -160,9 +160,8 @@ func handleAllocateRequest(req Request, stunMsg *stun.Message) error { //nolint:
 	_ = realmAttr.GetFrom(stunMsg)
 
 	// RFC 6156: Parse REQUESTED-ADDRESS-FAMILY attribute if present.
-	// If absent, default to IPv4 per RFC 6156 Section 4.1.1.
-	// "If the REQUESTED-ADDRESS-FAMILY attribute is absent, the server MUST
-	// allocate an IPv4-relayed transport address for the TURN client."
+	// If absent, default to the listener family unless strict RFC 6156
+	// behavior is requested.
 	// "Length:  this 16-bit field contains the length of the attribute in
 	// bytes.  The length of this attribute is 4 bytes."
 	// "If the server does not support the address family requested by the
@@ -173,7 +172,7 @@ func handleAllocateRequest(req Request, stunMsg *stun.Message) error { //nolint:
 	if err = requestedFamily.GetFrom(stunMsg); err != nil {
 		switch {
 		case errors.Is(err, stun.ErrAttributeNotFound):
-			requestedFamily = proto.RequestedFamilyIPv4
+			requestedFamily = defaultAllocationAddressFamily(req)
 		case stun.IsAttrSizeInvalid(err) || stun.IsAttrSizeOverflow(err):
 			return buildAndSendErr(req.Conn, req.SrcAddr, err, badRequestMsg...)
 		default:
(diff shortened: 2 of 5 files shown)


rg0now (approved):

rg0now:
I'm wondering how to improve the discoverability of the `StrictAddressFamily` flag. This is one of those hidden feature flags that can cause hours of debugging headaches. Or maybe LLMs solve this already? I often skip reading the subtle parts of the docs, they don't.

JoTurk:
> I'm wondering how to improve the discoverability of the StrictAddressFamily flag. This is one of those hidden feature flags that can cause hours of debugging headaches. Or maybe LLMs solve this already? I often skip reading the subtle parts of the docs, they don't.

But the question is, is this worse it to make this more discover-able? the idea of this flag if someone is trying to debug why we don't fail back to IPv4 or if someone made an issue, or if we want to switch the default behavior in the future if libwebrtc implements RFC-6156. But I don't think we should or can make every setting discoverable it might lead to cognitive overload.

PR #578, opened 2026-06-27
Close allocation on ChannelBind 400

### LINKED ISSUE #540: Coturn error: error 400: You cannot use the same peer with different channel number
In <https://github.com/jech/galene/issues/294>, a user is reporting that running Galene against Coturn causes connections to occasionally fail with the following in Coturn's log:

> error 400: You cannot use the same peer with different channel number

Since I'm running Galene with Pion's own TURN server, I haven't seen this error.  I might test against Coturn later next month, but if somebody has any insights, help would be welcome.

JoTurk:
#### Description
This tears down the allocation when  we receive a 400 Bad Request during channelbind, that makes it possible for the caller/ICE layer to recover by creating a fresh allocation after coturn or others rejects the bind with "You cannot use the same peer with different channel number"
#### Reference issue
should help resolve #540

> If the client receives a ChannelBind failure response that indicates that the channel information is out of sync between the client and the server (e.g., an unexpected 400 "Bad Request" response), then it is RECOMMENDED that the client immediately delete the allocation and start afresh with a new allocation.
> https://datatracker.ietf.org/doc/html/rfc8656

### DIFF
--- a/internal/client/udp_conn.go
+++ b/internal/client/udp_conn.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"math"
 	"net"
+	"sync"
 	"time"
 
 	"github.com/pion/stun/v3"
@@ -42,6 +43,7 @@ type UDPConn struct {
 	checkBindingsTimer     *PeriodicTimer    // Thread-safe
 	readCh                 chan *inboundData // Thread-safe
 	closeCh                chan struct{}     // Thread-safe
+	closeMutex             sync.Mutex        // Thread-safe
 	bindingRefreshInterval time.Duration     // Read-only
 	allocation
 }
@@ -188,6 +190,14 @@ func (c *UDPConn) WriteTo(payload []byte, addr net.Addr) (int, error) { //nolint
 	if !ok {
 		return 0, errUDPAddrCast
 	}
+	if c.isClosed() {
+		return 0, &net.OpError{
+			Op:   "write",
+			Net:  c.LocalAddr().Network(),
+			Addr: c.LocalAddr(),
+			Err:  errClosed,
+		}
+	}
 
 	// Check if we have a permission for the destination IP addr
 	perm, ok := c.permMap.find(addr)
@@ -276,6 +286,15 @@ func (c *UDPConn) Close() error {
 // LocalAddr returns the local network address.
 func (c *UDPConn) LocalAddr() net.Addr {
 	return c.relayedAddr
+}
+
+func (c *UDPConn) isClosed() bool {
+	select {
+	case <-c.closeCh:
+		return true
+	default:
+		return false
+	}
 }
 
 // SetDeadline sets the read and write deadlines associated
@@ -425,6 +444,9 @@ func (c *UDPConn) maybeBind(bound *binding) {
 		if err != nil {
 			c.log.Warnf("Failed to bind channel %d: %s", bound.number, err)
 			bound.setState(bindingStateFailed)
+			if errors.Is(err, errChannelBindBadRequest) {
+				c.closeAfterChannelBindBadRequest(bound)
+			}
 
 			return
 		}
@@ -452,6 +474,18 @@ func (c *UDPConn) maybeBind(bound *binding) {
 	go bind()
 }
 
+func (c *UDPConn) closeAfterChannelBindBadRequest(bound *binding) {
+	c.log.Warnf(
+		"ChannelBind rejected with 400 for %s on channel %d; closing TURN allocation",
+		bound.addr,
+		bound.number,
+	)
+
+	if err := c.Close(); err != nil && !errors.Is(err, errAlreadyClosed) {
+		c.log.Warnf("Failed to close TURN allocation after ChannelBind 400: %s", err)
+	}
+}
+
 func (c *UDPConn) bind(bound *binding) error {
 	setters := []stun.Setter{
 		stun.TransactionID,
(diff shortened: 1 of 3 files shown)


internal/client/udp_conn.go
```diff
@@ -278,6 +291,15 @@ func (c *UDPConn) LocalAddr() net.Addr {
...
 
+func (c *UDPConn) isClosed() bool {
+	select {
+	case <-c.closeCh:
+		return true
+	default:
+		return false
+	}
+}
+
```
jech, line 302: That's something we call on every write, and doing a non-blocking read takes the channel mutex.  What about making isClosed into an atomic.Bool?  That's both simpler and more efficient (unlike taking a mutex, an atomic read is essentially free, since it doesn't cause cacheline bouncing.)

internal/client/udp_conn.go
```diff
@@ -257,6 +267,9 @@ func (c *UDPConn) WriteTo(payload []byte, addr net.Addr) (int, error) { //nolint
 // Close closes the connection.
 // Any blocked ReadFrom or WriteTo operations will be unblocked and return errors.
 func (c *UDPConn) Close() error {
+	c.closeMutex.Lock()
+	defer c.closeMutex.Unlock()
+
```
jech, line 272: That's a separate fix, right?  Perhaps you'd want to split it into its own commit and push it to master straight away?

jech:
I've looked at the code, and I see nothing obviously wrong, except the bit about taking an extra mutex on every call.  See my comments inline.

Now, while I feel this is a good thing to do, I think it's a workaround to a problem that lies elsewhere.  I'd really like to understand the source of the problem.

internal/client/udp_conn.go
JoTurk: The go runtime does this for empty non blocked channels(this was changed a few years ago iirc, so while the the allocation is open this will use an atomic at the runtime, I'll link the runtime code / go issue when I'm not on my phone.

internal/client/udp_conn.go
JoTurk: It's a related fix because we didn't call close ourself, but now it can race because we and the user can call close.

JoTurk:
@jech 
>  I think it's a workaround to a problem that lies elsewhere. I'd really like to understand the source of the problem

I also have an idea to recover the binding state once we get 400 it will work for a few edge cases when the binding state is dropped at the wire, I'll open a pr once I'm on my pc.

internal/client/udp_conn.go
jech:
```
goos: linux
goarch: amd64
pkg: atomicbool
cpu: Intel(R) Core(TM) Ultra 7 265H
BenchmarkChannel-16    	306996789	         7.314 ns/op
BenchmarkAtomic-16     	1000000000	         0.3123 ns/op
PASS
```
Source code is here: <https://gist.github.com/jech/7e250f0fe599736bfe5ceb982f8040f8>

internal/client/udp_conn.go
jech: Still, why not move it into its own commit?

internal/client/udp_conn.go
JoTurk: This benchmark is incorrect, you measure your parallel code, and you don't measure against a mutex, this isn't even close to how slow a mutex is. Anyways your original comment was that it takes a channel mutex, this doesn't happen you can run this correct benchmark https://gist.github.com/JoTurk/b5e7d715e7f2ef9e31dc88c2ed6600fe

results are:
```
closed/channel   ~0.59-0.60 ns/op   0 allocs/op
closed/atomic    ~0.10 ns/op        0 allocs/op
closed/mutex     ~42-48 ns/op       0 allocs/op
closed/rwmutex   ~31 ns/op          0 allocs/op
```
and if you do `go test -run '^$' -bench '^BenchmarkCloseSignal/closed/channel$' -benchtime=1s -trace /tmp/turn-close-channel.trace test.go` and `go tool trace -pprof=sync /tmp/turn-close-channel.trace > /tmp/turn-close-channel-sync.pprof` you won't see a trace of a Mutex because go uses the fast path `selectnbrecv` check https://go.dev/src/runtime/chan.go

Channels are still slower than an atomic but they aren't a mutex slow, and we can't switch to a Atomic here because this will break this contract and will case a race:

> goroutine A: checks closed == false
> goroutine B: Close stores closed = true
> goroutine A: enters select and blocks on readCh/timer


so 0.60 ns/op vs 0.10 ns/op is a fair trade and it's not a Mutex (50ns/op slow).

> you do realise that this happens every time? Every time you ask me to review your code, I always end up having to do extra work because you won't take a comment until I've proven to you that I'm right

Sorry but this is childish, nobody is trying to win anything here, this is volunteer work and I do it for fun. Also you just moved the goal post from mutex to it's slower.

 > Joe
 
 Please don't call me joe. you used my correct name before.

internal/client/udp_conn.go
JoTurk: I'll break it into commits.

internal/client/udp_conn.go
jech: > we can't switch to a Atomic here because this will break this contract and will case a race

Ah, I missed the fact that closeCh wasn't introduced by this commit, it was already being used in a select elsewhere.  Point taken.

internal/client/udp_conn.go
JoTurk: done.

JoTurk:
@jech please check https://github.com/pion/turn/pull/579 when you have time, this hack will make it possible to recover the allocation if it got lost in the wire.

rg0now (approved):

PR #585, opened 2026-08-09
Add deterministic UDP peer preparation

the-sarge:
## Summary

Add `Client.PrepareUDPPeer(ctx, peer)`: an opt-in way for applications to wait until a UDP peer has a confirmed permission and ChannelBind before sending application packets.

This is a rework of the original diff. It now reuses and extends the existing binding manager, with the existing mutexes covering the bind, instead of introducing a new transaction/allocation lifecycle. The non-test diff is 447 insertions / 23 deletions (~320 code lines excluding comments and blanks), plus 340 lines of tests.

## Motivation

- Today the relayed conn sends via Send Indications and asynchronously upgrades a peer to ChannelData once the background ChannelBind completes. Applications cannot observe or control when that switch happens.
- The switch changes per-packet TURN overhead (~36 bytes → 4 bytes) mid-stream, so the maximum usable payload changes at a nondeterministic point in the connection's lifetime.
- This breaks packet-size-sensitive transports layered over the relayed conn — QUIC in particular, which pads Initials to 1200 bytes and validates path MTU once, then assumes it holds. An unobservable overhead change mid-handshake produces timing-dependent failures that are very hard to reproduce.

## Rework

Rebuilt per JoTurk's suggestion on Discord (make the existing mutex cover the bind; reuse/extend the current binds; no new transaction and allocation lifecycle):

- The previous +2668/−138 approach (new operation/lifecycle types, context-aware transaction plumbing) is dropped entirely.
- An in-flight ChannelBind attempt is now tracked by a per-binding channel under the existing `muBind`. `PrepareUDPPeer` waiters select on that channel, so same-peer callers coalesce onto one shared attempt and a caller's context cancellation wakes only that waiter with its cause, leaving the shared attempt running. The binding state machine itself is unchanged.
- Two small seams were added where the existing code had none: a `terminalize` transition folded into the binding's existing mutex, so a completing bind attempt cannot resurrect a permanently failed binding, and an `onPermRefreshFailure` callback on the allocation, because permission refresh failures were previously log-only but readiness must observe them.
- Deterministic close reuses the same primitives: `PeriodicTimer` gains `StopAndWait`, and bind/permission workers are tracked by a WaitGroup gated by the existing `closeMutex`.

## Proposed behavior

- `PrepareUDPPeer(ctx, peer)` resolves once the permission and ChannelBind for that peer are confirmed. After it succeeds, writes to that peer use ChannelData or return an error for the lifetime of the allocation — never a silent fallback to Send Indications — so the packet profile is deterministic before the first application byte.
- Concurrent calls for one peer coalesce on the existing binding state; different peers prepare independently. Caller cancellation wakes only that waiter with its exact cause and does not cancel shared allocation-owned permission/ChannelBind work.
- Peer addresses are reduced to a canonical form (IPv4-mapped IPv6 unmapped, zones handled), so an alias of a prepared peer shares its permission and channel binding rather than bypassing them via Send Indication.
- UDP allocation shutdown now joins refresh timers, binding checks, and in-flight ChannelBind/permission work before `Close` returns. Socket interruption remains the caller's responsibility: the client never closes the caller-owned `ClientConfig.Conn` or mutates its deadlines, so `Close` completes once the socket owner permits or actively unblocks in-flight I/O.
- Known bound: if the server stops responding, `Close` can take up to the STUN retransmission budget (~63.5 s with defaults) while an in-flight transaction runs out, because transaction waits are not interruptible in this minimal version. Prompt interruption (context-aware transaction waits) is deliberately left out to keep this PR small.

## Compatibility

- Purely additive for applications that never call `PrepareUDPPeer`, with one intentional refinement: `WriteTo` now canonicalizes `*net.UDPAddr` aliases of the same peer so they share one permission and one channel binding.
- Transaction and refresh plumbing is untouched; TCP allocations are untouched.

## Testing

- `go test -race -count=1 ./internal/client .` passes.
- `golangci-lint run` with the repo config reports no new findings versus `main`.
- New tests (all under `-race`): readiness success then ChannelData-only writes; invalid-peer rejection (unspecified, multicast, zoned, bad port, non-UDP); peer aliases sharing the prepared binding; same-peer coalescing onto one bind and one permission; waiter-local cancellation with its cause while the shared bind survives; permission refresh failure failing writes with no Send-indication fallback; bind failure surfacing to the preparing caller; terminal failure surviving an in-flight bind success; `Close` joining in-flight bind workers.

### DIFF
--- a/client.go
+++ b/client.go
@@ -4,6 +4,7 @@
 package turn
 
 import (
+	"context"
 	b64 "encoding/base64"
 	"fmt"
 	"math"
@@ -569,6 +570,21 @@ func (c *Client) CreatePermission(addrs ...net.Addr) error {
 	}
 
 	return nil
+}
+
+// PrepareUDPPeer creates a permission for peer on the client's UDP allocation
+// and waits until the TURN server confirms a channel binding for it. After it
+// returns nil, writes to peer use ChannelData (or fail) for the lifetime of
+// the allocation; they never fall back to Send indications. Concurrent calls
+// for the same peer share one permission and one bind; canceling ctx wakes
+// only that caller and leaves the shared work running.
+func (c *Client) PrepareUDPPeer(ctx context.Context, peer net.Addr) error {
+	conn := c.relayedUDPConn()
+	if conn == nil {
+		return errUDPAllocationNotFound
+	}
+
+	return conn.PreparePeer(ctx, peer)
 }
 
 // PerformTransaction performs STUN transaction.
--- a/errors.go
+++ b/errors.go
@@ -24,6 +24,7 @@ var (
 	errSTUNServerAddressNotSet       = errors.New("STUN server address is not set for the client")
 	errOneAllocateOnly               = errors.New("only one Allocate() caller is allowed")
 	errAlreadyAllocated              = errors.New("already allocated")
+	errUDPAllocationNotFound         = errors.New("UDP allocation not found")
 	errNonSTUNMessage                = errors.New("non-STUN message from STUN server")
 	errFailedToDecodeSTUN            = errors.New("failed to decode STUN message")
 	errUnexpectedSTUNRequestMessage  = errors.New("unexpected STUN request message")
--- a/internal/client/allocation.go
+++ b/internal/client/allocation.go
@@ -49,6 +49,10 @@ type allocation struct {
 	readTimer         *time.Timer           // Thread-safe
 	mutex             sync.RWMutex          // Thread-safe
 	log               logging.LeveledLogger // Read-only
+
+	// onPermRefreshFailure, when set, observes a permission refresh that kept
+	// failing after retries. Read-only after construction.
+	onPermRefreshFailure func(error)
 }
 
 func (a *allocation) setNonceFromMsg(msg *stun.Message) {
@@ -166,6 +170,9 @@ func (a *allocation) onRefreshTimers(id int) {
 		}
 		if err != nil {
 			a.log.Warnf("Failed to refresh permissions: %s", err)
+			if a.onPermRefreshFailure != nil {
+				a.onPermRefreshFailure(err)
+			}
 		}
 	}
 }
--- a/internal/client/binding.go
+++ b/internal/client/binding.go
@@ -37,12 +37,52 @@ type binding struct {
 	addr         net.Addr        // Read-only
 	mgr          *bindingManager // Read-only
 	muBind       sync.Mutex      // Thread-safe, for ChannelBind ops
+	attemptDone  chan struct{}   // Protected by muBind; non-nil while a bind attempt is in flight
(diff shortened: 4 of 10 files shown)


the-sarge:
@JoTurk reworked per the Discord discussion — thanks for the pointer. The original approach is dropped entirely; the branch now reuses and extends the current binds with the existing mutex covering the bind, and no new transaction or allocation lifecycle.

Where it landed: ~447 non-test lines including comments (~320 code lines). That covers both readiness (`PrepareUDPPeer`: permission + server-confirmed ChannelBind, same-peer callers coalesce, cancellation wakes only the canceled waiter, no silent Send-indication fallback afterwards) and deterministic close (timers and bind/permission workers joined before `Close` returns).

Tests, all under `-race`: readiness → ChannelData-only writes, peer-alias canonicalization, same-peer coalescing, waiter-local cancellation, permission-refresh failure fails writes (never Send indication), sticky terminal failure against an in-flight bind success, and Close joining in-flight workers. Happy to adjust further.

the-sarge:
Heads up — while poking at the close wait-time gap I mentioned on Discord, I found a bug in this branch: a second `PrepareUDPPeer` call for the same peer could block uncancellably while the first caller's CreatePermission transaction was in flight (it waited on the permission mutex instead of the attempt channel). Fixed by moving attempt bookkeeping onto its own mutex, with a regression test using a delayed permission transaction. +12 lines, one test.

While I was in there I started putting together that wait-time follow-up for after this lands. Shape so far:
- ~85 non-test lines (plus 216 of tests) on top of this PR
- zero public API change — the primitive already exists (`Transaction.Close()`, same thing `Client.Close()` uses); the follow-up just wires an allocation's close to abort its own in-flight exchanges, so `Close()` simply gets fast
- measured, not estimated: against a server that goes silent, close latency drops from the full retransmission budget (~63.5 s at defaults) to about a tenth of a millisecond

the-sarge:
The macOS 1.25 failure is a flake in the pre-existing `TestPeriodicTimer/stop_inside_handler` subtest — it sleeps 30ms against a 20ms timer, so a slow runner can run the assert before the handler has fired (reproduced that exact failure signature locally by tightening the margin; a fired handler that calls Stop provably can't leave the timer running). Pushed a small commit making the subtest event-driven instead of sleep-calibrated — timer behavior unchanged.

the-sarge:
@JoTurk let me know if you'd rather I squash/force these.

JoTurk:
@the-sarge It's fine to keep the commits we can always squash merge or do it before merge.
Sorry i got more busy with the DTLS 1.3 project, I'll try to review this tomorrow. Thank you for the review and the fixes.

the-sarge:
No worries - I had no idea how many different projects you guys had going on! Impressive. Thanks for all of it.

the-sarge:
I decided to go a different way so I could make progress...feel free to close this if you don't want it.

PR #587, opened 2026-08-15
Fix deadlock on duplicate CreateTCPConnection

the-sarge:
## Summary

On the duplicate-connection path, `Manager.CreateTCPConnection` returned `ErrDupeTCPConnection` while still holding `m.lock`:

```go
m.lock.Lock()
if m.isDupeTCPConnection(allocation, remoteAddr) {
    return 0, ErrDupeTCPConnection // lock never released
}
m.lock.Unlock()
```

Once that path is hit, the allocation manager's mutex is held forever, so every subsequent operation that needs it (new allocations, lookups, deletions, `Close`) blocks permanently. Since an authenticated client can trigger it simply by issuing a second RFC 6062 Connect request for the same peer address, this is a remotely triggerable denial of service that wedges the whole server.

The fix evaluates the dupe check under the lock, releases it, and then returns.

## Test

The existing `TestCreateTCPConnectionDuplicateTCPConn` passed even with the deadlock present, because it never touched the manager lock after the duplicate rejection. It is rewritten as `TestCreateTCPConnectionDuplicateDoesNotBlockManager`, which asserts that `AllocationCount()` (which takes the lock) returns promptly after a duplicate `Connect` is rejected — with the old code this times out.

## Context

This was reported privately by email; opening this PR at the maintainer's request.

### DIFF
--- a/internal/allocation/allocation_manager.go
+++ b/internal/allocation/allocation_manager.go
@@ -377,10 +377,11 @@ func (m *Manager) CreateTCPConnection( // nolint: cyclop
 	remoteAddr := &net.TCPAddr{IP: peerAddress.IP, Port: peerAddress.Port}
 
 	m.lock.Lock()
-	if m.isDupeTCPConnection(allocation, remoteAddr) {
+	isDupe := m.isDupeTCPConnection(allocation, remoteAddr)
+	m.lock.Unlock()
+	if isDupe {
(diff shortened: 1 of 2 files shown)


JoTurk (approved):
Thank you

the-sarge:
fixed the title for your linter

PR #592, opened 2026-08-24
Add a channel number lookup and an active-channel check

rg0now:
This PR proposes a simpler alternative to the `PrepareUDPPeer` client API proposed in #585.

#### Description

PR #585 adds `PrepareUDPPeer(ctx, peer)` so an application can block until a peer has a confirmed permission and bound a channel before sending (this is useful for QUIC senders). This PR exposes the channel state instead of managing it, which allows a caller to poll the channel's active state to know when the channel comes online (and when it goes away).

Added is a new `IsChannelActive(chNum uint16) bool` client API that reports the same condition `WriteTo` uses to choose ChannelData over a SendIndication. The PR also adds `FindChannelNumberByAddr`, the counterpart to the existing `FindAddrByChannelNumber`, so a caller holding a peer address can learn the channel number. Tests now cover all three accessors across every binding state.

#### Reference

#585


JoTurk:
@rg0now sorry i missed this, just saw it in my notification, I'll review and test after my work tonight

rg0now:
It's somewhat disillusioning that I can't get a PR right without AI for the first try. Not even for the second. Copilot caught some real bugs here. 

Remaining two issues flagged Copilot, although valid, are beyond the scope of this PR so I flagged them WONTFIX. When we'll have ample time (never) we may want to return to them

JoTurk (approved):
Sorry for missing this, looks good and works as intended

PR #593, opened 2026-08-27
Fix uint16 overflow in consumeSingleTURNFrame

GerardGao:
SUMMARY

A malformed ChannelData frame with a length field in [0xFFFC, 0xFFFF]
caused the frame-size computation in consumeSingleTURNFrame to
overflow uint16, wrapping to 0. STUNConn.ReadFrom then reported a
zero-size frame with a nil error and never consumed the buffered
data, leaving the server read loop spinning at 100% CPU. A single
4-byte packet over TCP triggers this, allowing a remote DoS.

The frame size is now computed with uint32 arithmetic, and the
padding math was rewritten in an equivalent overflow-safe form.
The same overflow class in the STUN branch is covered by the fix.

TEST

- TestOverflowRepro_ConsumeFrame / TestOverflowRepro_ReadFromDoesNotConsume
  fail before the fix (n=0, err=nil) and pass after.
- TestPaddingFormulaEquivalence exhaustively proves the new padding
  formula matches the old one for all 65536 length values.
- Boundary scans cover 0xFFF8-0xFFFF, STUN-length overflow, max frame
  size (65540), zero-length and non-multiple padding.
- golangci-lint v2.10.1: 0 issues.
- go test -race on internal/proto, internal/client, internal/server: pass.

CONTEXT

Found while reviewing the codebase for potential protocol-handling
bugs. The fix is limited to internal/proto and changes no public API.


fallenmi (changes requested):
Disclosure: I used Codex to assist with this review and verified the result against exact commit `c093d9d52cf9b01f30f7ddee5e911091d50688ff`.

Blocking: `STUNConn.ReadFrom` can now return a frame size larger than the destination buffer. At the boundary fixed here, a ChannelData length of `0xFFFC` produces a 65,536-byte frame, while the standard TCP client allocates a 65,535-byte buffer. `copy(payload, s.buff[:n])` copies only 65,535 bytes, but the function returns `n=65536`; `Client.Listen` then slices `buf[:n]` and panics.

I reproduced this on both the exact head and current GitHub merge with the documented `NewSTUNConn(net.Conn)` path:

```text
panic: runtime error: slice bounds out of range [:65536] with capacity 65535
github.com/pion/turn/v5.(*Client).Listen.func1()
    client.go:300
```

The exhaustive arithmetic oracle itself is clean on this head (all 65,536 ChannelData lengths and all 16,384 legal STUN lengths match the 32-bit reference), so the widening is correct. Please also make `ReadFrom` handle `n > len(payload)` without reporting more bytes than were copied, and add an end-to-end boundary test through `Client.Listen`; otherwise the original high-length DoS is replaced by a client crash for a complete frame.

GerardGao:
@fallenmi Thanks for the thorough review. Both points are addressed in the latest push (ca3db96):

1. ReadFrom no longer reports more bytes than copied: when a complete frame is larger than the caller's buffer, ReadFrom consumes the entire frame from the stream (keeping framing aligned) but returns only the bytes actually copied, per the net.PacketConn contract. Client.Listen no longer panics on a 65,536-byte ChannelData frame.
2. Added end-to-end boundary test: TestClientListenWithOversizedChannelData reproduces the exact "slice bounds out of range [:65536] with capacity 65535" panic against the previous head and passes now. Also added TestReadFromOversizedFrameReportsCopiedBytes and TestReadFromContinuesAfterOversizedFrame covering the return-value contract and stream alignment.

Verified locally: proto and root package -race suites pass, golangci-lint reports 0 issues, and the existing overflow/boundary tests are unchanged.

fallenmi (approved):
I re-reviewed exact changed head `ca3db96f5ca38bb424cae4931bc7436b11b169a9`.

The prior buffer-contract blocker is fixed: `STUNConn.ReadFrom` now consumes the complete oversized frame to keep stream framing aligned while returning only the bytes actually copied. The new boundary and next-frame tests pass ten race-enabled repetitions on both this head and the current merge.

I also ran an independent oracle through the actual `Client.Listen()` method, rather than a mirrored loop. After a legal 65,536-byte ChannelData frame is truncated to the client's 65,535-byte buffer, the listener remains active and proceeds to its next read; this passed five race-enabled repetitions on both head and merge. The old slice-bounds panic is gone without corrupting the following frame.

All 18 live checks, eight check suites, and six workflows are successful. I found no remaining blocker.

Disclosure: I used OpenAI Codex to re-review the changed head, inspect the exact fix and live repository state, and run the exact-head/current-merge race oracles described above; I independently verified the result.

JoTurk (changes requested):
The 65536-byte client-crash is unrealistic, Turn server buffer is limited to 1600-bytes and pion/webrtc has a 1200 mtu. 

this PR should be only limited to safe frame-length handling, and making sure we don't overflow, and we should reject frames whose declared size exceeds len(payload)

GerardGao:
@JoTurk agreed on both points — the PR is now narrowed to exactly what you described. Pushed in `0b539fa`.

**Dropped the client-crash framing.** You're right that a 65,536-byte frame is not reachable through the pion stack (server `inboundMTU`, webrtc's 1200 MTU). The oversized-frame handling is no longer justified as a DoS fix, so I removed the truncate-and-deliver path and the `Client.Listen` end-to-end test built around it.

**What remains, per your direction:**

1. *Don't overflow.* Frame size is computed in `uint32`. On `main`, a ChannelData length of `0xFFFE`/`0xFFFF` wraps and `consumeSingleTURNFrame` returns `n=4, err=nil` — a "valid" 4-byte frame for a declared 65 KB one, so `ReadFrom` reports a frame it never consumed and the read loop spins. Verified RED on `main` / GREEN on this head.

2. *Reject frames whose declared size exceeds `len(payload)`.*

```go
if n > len(payload) {
    s.buff = s.buff[n:]
    return 0, nil, errTURNFrameTooLarge
}
```

The frame is consumed so stream framing stays aligned and the next frame still parses; no caller ever observes `n > len(payload)`.

**Tests:** folded into the existing `internal/proto/stun_conn_test.go` — no separate regression files. Three additions: the `0xFFF8..0xFFFF` / STUN-length overflow scan, an exhaustive proof that the rewritten padding formula matches the original for all 65,536 length values, and a `ReadFrom` case asserting rejection plus clean parsing of the following frame.

**Diff is now 2 files, +112/-11**, confined to `internal/proto`, no public API change.

Verified: `go build ./...`, `go vet ./...`, `go test -race ./internal/proto/...` pass; `golangci-lint run` reports zero issues in the changed files. (`TestCreateTCPConnectionInvalid` and `TestConnectRequest` fail identically on unmodified `main` in my environment — pre-existing, unrelated.)

GerardGao:
Rebased the three commits to use a consistent author identity (`GerardGao <213731635+GerardGao@users.noreply.github.com>` for all three — one had a corporate email). No code change at all; the diff is identical to the previous push.

To summarise what the current head contains, per your direction:

1. **Overflow fix only.** `datagramSize` is now `uint32`; the old `uint16` arithmetic wrapped on `dataLen >= 0xFFFC`, producing a near-zero frame size that made `ReadFrom` return `(0, nil)` without consuming the buffer and spin forever.
2. **Reject frames whose declared size exceeds `len(payload)`.** When `n > len(payload)`, `ReadFrom` advances the buffer by the full frame size (stream framing stays aligned) and returns `errTURNFrameTooLarge`. No truncate-and-deliver path.
3. **No client-crash test.** The `Client.Listen` end-to-end test and `client_oversized_frame_test.go` are gone. Tests are limited to `internal/proto`: an overflow scan for `0xFFF8..0xFFFF`, an exhaustive padding-formula equivalence proof, and a `ReadFrom` boundary test with a 1600-byte payload (matching the realistic server buffer).

Could someone approve the workflow run? CI has not triggered yet — I believe it needs a maintainer to approve since this is my first contribution here.

GerardGao:
Hi @JoTurk, I've addressed all your review feedback:

- Removed the unrealistic client-crash test scenario (large frame exceeding 64 KB buffer)
- Kept the fix scope to two things as you requested: safe frame-length handling via `uint32` (preventing `uint16` overflow) and rejecting frames whose declared size exceeds the available payload
- Frames that are too large are now consumed in full to maintain stream alignment before returning `errTURNFrameTooLarge`

Could you please take another look and, if everything looks good, approve the CI workflow run? (First-contribution restriction requires maintainer approval to trigger GitHub Actions.)

Thank you!

PR #594, opened 2026-08-28
Reject cross-family peers in Send indications

### LINKED ISSUE #566: RFC 6156 default-IPv4 family blocks TURN on IPv6-only Kubernetes clusters

GerardGao:
SUMMARY

RFC 6156 constrains every peer of an allocation to the address family
of its relayed transport address. The server enforces this for
CreatePermission and ChannelBind requests, but handleSendIndication
historically relayed data whenever a matching permission existed,
even when the peer address family differed from the allocation's.

A Send indication carries no response, so mismatched data is
discarded (logged and dropped) instead of rejected with a 443 error,
consistent with how RFC 5766 handles data without a permission.

TEST

- TestHandleSendIndicationAddressFamilyMismatch fails before the fix
  (cross-family data was written to the relay socket with no error)
  and passes after.
- TestHandleSendIndicationSameFamilyStillRelays guards against
  over-rejection.
- Existing TestHandleSendIndication subtests, CreatePermission and
  ChannelBind family tests all pass.
- golangci-lint v2.10.1: 0 issues. go test -race: pass.

CONTEXT

This complements the family-consistency model established by #575
(which closed #566): rather than relaxing the family checks, #575
fixed the default allocation family. This change applies the same
model to the remaining handler that was missing it.

Note: PR #567 proposes relaxing ipMatchesFamily for dual-stack
relays; if adopted, the Send indication check added here would need
the same relaxation.

### DIFF
--- a/internal/server/turn.go
+++ b/internal/server/turn.go
@@ -469,6 +469,15 @@ func handleSendIndication(req Request, stunMsg *stun.Message) error {
 		return err
 	}
 
+	// RFC 6156: Peer address must match allocation's address family.
+	// Unlike CreatePermission and ChannelBind, a Send indication carries no
+	// response, so mismatched data is discarded instead of rejected with 443.
+	if !ipMatchesFamily(peerAddress.IP, alloc.AddressFamily()) {
+		req.Log.Infof("peer address family mismatch for client %s to peer %s", req.SrcAddr, peerAddress.IP)
+
+		return errPeerAddressFamilyMismatch
+	}
+
 	msgDst := &net.UDPAddr{IP: peerAddress.IP, Port: peerAddress.Port}
 	if perm := alloc.GetPermission(msgDst); perm == nil {
 		return fmt.Errorf("%w: %v", errNoPermission, msgDst)
--- a/internal/server/turn.go
+++ b/internal/server/turn.go
@@ -469,18 +469,22 @@ func handleSendIndication(req Request, stunMsg *stun.Message) error {
 		return err
 	}
 
-	// RFC 6156: Peer address must match allocation's address family.
-	// Unlike CreatePermission and ChannelBind, a Send indication carries no
-	// response, so mismatched data is discarded instead of rejected with 443.
-	if !ipMatchesFamily(peerAddress.IP, alloc.AddressFamily()) {
-		req.Log.Infof("peer address family mismatch for client %s to peer %s", req.SrcAddr, peerAddress.IP)
-
-		return errPeerAddressFamilyMismatch
-	}
-
 	msgDst := &net.UDPAddr{IP: peerAddress.IP, Port: peerAddress.Port}
+	// RFC 8656 Section 11.2: the permission check comes before any
+	// address/port restrictions, and a Send indication never refreshes the
+	// permission.
 	if perm := alloc.GetPermission(msgDst); perm == nil {
 		return fmt.Errorf("%w: %v", errNoPermission, msgDst)
+	}
+
+	// RFC 8656 Section 11.2: the server MAY impose restrictions on the peer
+	// address; the peer address family must match the allocation's address
+	// family (see Section 10.2). A Send indication carries no response, so
+	// mismatched data is silently discarded.
+	if !ipMatchesFamily(peerAddress.IP, alloc.AddressFamily()) {
+		req.Log.Debugf("peer address family mismatch for client %s to peer %s", req.SrcAddr, peerAddress.IP)
+
+		return errPeerAddressFamilyMismatch
 	}
 
 	l, err := alloc.WriteTo(dataAttr, msgDst)
--- a/internal/server/turn_family_test.go
+++ b/internal/server/turn_family_test.go
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: 2026 The Pion community <https://pion.ly>
 // SPDX-License-Identifier: MIT
 
+//go:build !js
+
 package server
 
 import (
@@ -51,7 +53,7 @@ func (c *blockingRelayConn) Close() error {
 
 // TestHandleSendIndicationAddressFamilyMismatch verifies that a Send
 // indication addressed to a peer of a different address family than the
-// allocation is not relayed, matching the RFC 6156 model where the relayed
+// allocation is not relayed, matching the RFC 8656 model where the relayed
 // transport address family constrains every peer it communicates with.
 //
 // CreatePermission and ChannelBind already enforce this (443 Peer Address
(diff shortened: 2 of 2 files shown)


internal/server/turn.go
```diff
@@ -469,6 +469,15 @@ func handleSendIndication(req Request, stunMsg *stun.Message) error {
...
 
+	// RFC 6156: Peer address must match allocation's address family.
+	// Unlike CreatePermission and ChannelBind, a Send indication carries no
+	// response, so mismatched data is discarded instead of rejected with 443.
+	if !ipMatchesFamily(peerAddress.IP, alloc.AddressFamily()) {
+		req.Log.Infof("peer address family mismatch for client %s to peer %s", req.SrcAddr, peerAddress.IP)
+
+		return errPeerAddressFamilyMismatch
+	}
+
```
JoTurk: Per rfc-8656 permission checks must be done before any address / port checks.

internal/server/turn.go
```diff
@@ -469,6 +469,15 @@ func handleSendIndication(req Request, stunMsg *stun.Message) error {
 		return err
 	}
 
+	// RFC 6156: Peer address must match allocation's address family.
+	// Unlike CreatePermission and ChannelBind, a Send indication carries no
+	// response, so mismatched data is discarded instead of rejected with 443.
+	if !ipMatchesFamily(peerAddress.IP, alloc.AddressFamily()) {
+		req.Log.Infof("peer address family mismatch for client %s to peer %s", req.SrcAddr, peerAddress.IP)
```
JoTurk: this log should be changed to debug or removed (better).

JoTurk (changes requested):

internal/server/turn.go
GerardGao: Fixed: the permission check now runs before the address family restriction, per RFC 8656 Section 11.2. A cross-family peer without a permission is now dropped as errNoPermission. Added TestHandleSendIndicationPermissionBeforeFamily to lock the ordering.

internal/server/turn.go
GerardGao: Fixed: the log is now Debugf to avoid flooding the log on the data path. The control-plane handlers keep Infof.

GerardGao:
Thanks for the review. Both issues are fixed in the latest push:

1. The permission check now runs before the address family restriction, per RFC 8656 Section 11.2 (a cross-family peer without a permission is dropped as errNoPermission). Added TestHandleSendIndicationPermissionBeforeFamily to lock the ordering.
2. The mismatch log is now Debugf (the control-plane handlers keep Infof).
3. Also added the missing //go:build !js tag to the test file, which was breaking the WASM CI job.

RFC references in the code and tests were updated from RFC 6156 to RFC 8656, the current specification.

GerardGao:
Rebased the two commits to use a consistent author name (`GerardGao` instead of `Gerard`) — no code change, diff is unchanged (`internal/server/turn.go +13`, `internal/server/turn_family_test.go +220`).

Both of JoTurk's requests from the original review are in place:
1. Permission check runs before the address-family check (RFC 8656 §11.2 ordering) — locked by `TestHandleSendIndicationPermissionBeforeFamily`.
2. The mismatch log is `Debugf`.

I think no CI has run on this branch yet because it needs a maintainer to approve the workflow run — this is my first contribution to this repo. Could someone approve it? Happy to address any further feedback.

GerardGao:
Hi @JoTurk, I've addressed all your review feedback:

- Moved the permission check before the address-family check, per RFC 8656 §11.2
- Changed the log from `Infof` to `Debugf` to avoid flooding the data-path log
- Updated the RFC reference from 6156 to 8656
- Added `//go:build !js` tag to fix the WASM CI failure
- Added `TestHandleSendIndicationPermissionBeforeFamily` to lock down the check ordering

Could you please take another look and, if everything looks good, approve the CI workflow run? (First-contribution restriction requires maintainer approval to trigger GitHub Actions.)

Thank you!

PR #598, opened 2026-09-07
Add AllocateWithContext to Client

Sean-Der:
Currently we don't have a clean way to cancel an Allocation and in pion/ice just block until complete. This gives us a cleaner/quicker way to cleanup.

Relates to pion/ice#781


rg0now:
Sorry for the late comment, but wouldn't it be more idiomatic Go to make the context the first argument of `PerformTransactionWithContext` rather than the last?

Sean-Der:
Yes! Sorry about that @rg0now 

I'll switch it real quick

